PR #18, opened 2014-09-12
allow fetching current data, and passing in old data to be merged with c...

wheybags:
...urrent data when generating reports

The idea is to able to support saving the current data set to a json file, like so:
```
data = JSON.generate Coverband::Reporter.get_current_scov_data
File.write("blah.json", data)
```

Then later on, pass it in to the html reporter:
```
data = JSON.parse(File.read("blah.json"))
Coverband::Reporter.report :additional_scov_data => [data]
```

### DIFF
--- a/lib/coverband/reporter.rb
+++ b/lib/coverband/reporter.rb
@@ -24,18 +24,18 @@ module Coverband
         return
       end
       redis = Coverband.configuration.redis
-      roots = Coverband.configuration.root_paths
+      roots = get_roots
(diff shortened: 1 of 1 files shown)


danmayer:
new version released 1.0.2 includes your change, another waiting PR, and some fixes related to recording the baseline in Sinatra apps.

danmayer:
thanks for the PR, I was out of the country for months and on my return took me awhile to get caught up with everything

PR #22, opened 2015-03-12
Use `count` instead of `increment` to measure this

strzalek:
I think we should be using `count` here instead of `increment`. Increment takes `step` as a second value, not the actual value, In fact, count is used by `increment` internally (https://github.com/reinh/statsd/blob/5d1d539e447053d28003c9b8f211dd5a35007dcd/lib/statsd.rb#L279).

It also makes it incompatible with some statsd clients (they take hash as a second argument).

### DIFF
--- a/lib/coverband/base.rb
+++ b/lib/coverband/base.rb
@@ -89,7 +89,7 @@ module Coverband
       if @reporter
         if @reporter.class.name.match(/redis/i)
           before_time = Time.now
-          @stats.increment "coverband.files.recorded_files", @files.length if @stats
+          @stats.count "coverband.files.recorded_files", @files.length if @stats
           @reporter.store_report(@files.dup)
           time_spent = Time.now - before_time
           @stats.timing "coverband.files.recorded_time", time_spent if @stats


ggilder:
:+1:

danmayer:
cool thanks. I think you are correct. I am moving tomorrow. So give me a couple days to get to this. I haven't dug into coverband to much lately. I have been meaning to get back to it, there are a number of cool improvements I could make.

strzalek:
@danmayer ping, a friendly reminder about the PR :)

danmayer:
Yep sorry been a bit slammed. I think I should have time to look at this tomorrow. Thanks for the ping ;)

strzalek:
Thanks Dan! :)

danmayer:
No worries, working on getting a new version released shortly.

danmayer:
new version released 1.0.2 includes your change, another long waiting PR, and some fixes related to recording the baseline in Sinatra apps.

strzalek:
Nice. Thanks!

PR #34, opened 2016-01-05
Filter files before reporting

kbaum:
Simply adding files to hash and waiting until filtering at the end reduces the amount of filtering we need to do.

Performance benchmark before:

```
coverband        14.560000   0.070000  14.630000 ( 14.674790)
no coverband      0.190000   0.000000   0.190000 (  0.193216)
```

After
```
coverband         7.710000   0.090000   7.800000 (  7.817221)
no coverband      0.170000   0.000000   0.170000 (  0.169859)
```


danmayer:
same looks good to merge, but needs a cleanup.

kbaum:
This has been pulled in with other pull requests

PR #35, opened 2016-01-05
Slightly improved performance on file check

kbaum:
Changed 

image

Using @ignore_patterns for hard coded ignores makes the code a little simpler.  Using just include instead of string regexp should speed things up a bit:

before:

```
~/tmp/coverband $ rake benchmarks
coverband        25.880000   0.170000  26.050000 ( 26.983449)
no coverband      0.810000   0.060000   0.870000 (  0.973907)
```

after:

```
~/tmp/coverband $ rake benchmarks
coverband        23.180000   0.210000  23.390000 ( 23.618931)
no coverband      1.140000   0.050000   1.190000 (  1.218236)
```


danmayer:
So this one is a bit trickier, I have seen cases where the regex is actually really handy and using full patterns. I get that include is faster, but not sure we would want to dump functionality. Wondering if we could detect or allow both using include for most and regex as a second slower set.

kbaum:
Maybe we should add a new config option:

```
Coverband.configuration.ignore_regexp
```

danmayer:
:+1: to this idea `Coverband.configuration.ignore_regexp`

PR #36, opened 2016-01-05
Trace point for ruby 2 plus

kbaum:
Using TracePoint for ruby >= 2 gives a massive speed improvement.

Before:

```
coverband         6.110000   0.050000   6.160000 (  6.171213)
no coverband      0.170000   0.000000   0.170000 (  0.171993)
```

After:

```
coverband         0.840000   0.010000   0.850000 (  0.852229)
no coverband      0.160000   0.000000   0.160000 (  0.168215)
```


kbaum:
@danmayer - the builds seemed to be sporadically failing due to a bundler problem but i'm not sure what's happening:

image

kbaum:
@danmayer - fixed travis build failure.

danmayer:
Nice, yeah it would be kind of nice to keep Ruby 1.9 support but really tracepoint is getting much better and to get this project fully where we would want it it is likely best to just drop support. I think might be good to just add a note in the readme that version X is the last version that will support Ruby 1.9

kbaum:
Definitely agree. I think we should also note that the latest doesn't need
or use the c extension at this point.

danmayer:
Agreed, should cleanup the docs, perhaps we do a major version bump on the next gem release to signify dropping support and making some significant changes / not requiring the c_extension.

kbaum:
Sounds like a plan!

PR #39, opened 2016-01-07
added ensure to middleware to ensure tracing is stopped in this case

kbaum:
Thinking it makes sense to ensure that each requests stops the sampling in the case of an error being raised by the app.


danmayer:
cool just need to clean up merge

kbaum:
@danmayer - merged cleaned up.  Thx!

PR #45, opened 2016-01-20
Optional gem ignore

dnasseri:
This PR adds an include_gems configuration option so that we can be explicit about whether we want to ignore gems or not.


dnasseri:
cc @kbaum

kbaum:
Looks good but this test is failing for me:

image

Is it failing for you?

thx!

Issue #42, opened 2016-01-12
Sharing coverage report to the rest of the team


kbaum:
I've noticed that actually generating and sharing the coverage report to the rest of the team can be a bit awkward.  Code coverage data is stored within our production redis.  In order to generate the report, i have to copy that data to my local redis instance.  Then i run the `coverband:coverage` task locally.  This is a bit confusing as the path to the files on my local instance is different than in production.  It seems to work but looking at the code, i think this may not work for others depending on their coverband settings.  

Once i have the coverband report locally generated, now i have to figure out how to share it.  The actual coverage data is quite sensitive data given that it has the details of our codebase.  

I am not sure of the exact solution to the problem, but i think it would be neat if we could generate and expose the report within production.  Since we are on Heroku, we cannot generate it to the local file system.  We could generate and store on s3 or another remote file system.

Curious if others have noticed this issue.

thx!

danmayer:
So an approach similar to sidekiq web, flipper-ui, or others could work. Where we make a mountable tiny web frontend. Often just a single endpoint mountable Sintra works. I think in our case we would just have to allow the report to be generated in that front end and perhaps write the html report to redis or load the html templates that generate the report and pull them on the fly. 

https://github.com/jnunemaker/flipper/blob/master/lib/flipper/ui.rb

A thought at least

kbaum:
@danmayer - i think we use the exact same internal tools.  Love flipper! :wink: 

I am thinking the same thing.  The problem is that the report generation tends to take a while.  We could pre-generate it with a rake task and then store it in redis like you said.

kbaum:
@danmayer - spiked on this:

https://github.com/danmayer/coverband/compare/write_output_to_s3?w=0

Basically writes the coverage report to s3.  We could then create a little sinatra app that reads from that bucket and displays the results.

danmayer:
nice +1

PR #51, opened 2016-02-02
Read/Write output to s3

kbaum:
This pull request allows for generating and reading coverage results within production.  

1. Simplifies sharing results with others on the team.  https://github.com/danmayer/coverband/issues/42
2. Avoids any complications with generating coverage data within production and generating report within dev

When the s3_bucket configuration is set, coverband will write to the ```coverband/index.html``` object within the s3_bucket.

```ruby
config.s3_bucket = 'yourapp-coverage-bucket'
```

Then mount the ```Coverband::S3Web``` sinatra app within your rails routes.rb.

```ruby
require 'coverband/s3_web'

YourApp::Application.routes.draw do
  authenticate :user, lambda { |u| u.admin? } do
    mount Coverband::S3Web => '/coverband/'
  end
end
```

Now when you run the rake coverband:coverage task, you'll need to ensure that the aws creds are set according the instructions within [aws-sdk](http://docs.aws.amazon.com/sdkforruby/api/index.html).

```bash
AWS_REGION=us-east-1 AWS_ACCESS_KEY_ID=1kdfslkdkd AWS_SECRET_ACCESS_KEY=dfkas93 rake coverband:coverage
```

### DIFF
--- a/coverband.gemspec
+++ b/coverband.gemspec
@@ -26,4 +26,5 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "simplecov"
   spec.add_runtime_dependency "json"
   spec.add_runtime_dependency "redis"
+  spec.add_runtime_dependency 'aws-sdk', '~> 2'
 end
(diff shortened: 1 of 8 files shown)


danmayer:
nice

PR #64, opened 2016-07-14
Write with additional data

inoda:


danmayer:
cool code looks good, and documentation is great... Could you explain a bit more about why you want this feature why write it out opposed to the redis store?

inoda:
@danmayer Sure. So on each deploy, I want to write the current coverage to a file, clear the Redis store, and re-take a baseline (clearing the store and re-taking the baseline is what your docs suggest if I don't want to muddle line numbers).

After the new baseline is taken, I want to merge the last coverage data for reporting. That works fine when I want to view the coverage. But the next time I want to write to a file, the last coverage data isn't being included. So this code allows for me to keep the file I'm writing to up to date with past coverage.

inoda:
@danmayer ping

danmayer:
so I can merge this in the issue with writing out the data and re-applying it is the same as re-using the data in redis... if you change the code the stored data would have wrong line counts... So it is a bit up to the user to see how often to do that... after small deploys I didn't tend to clear the redis, but on major changes or every few weeks I would... This would basically push that to the ability to keep and merge files but you still need to be cautious that line numbers match the time the code was live... So I will merge but we probably need to clarify how that all works in the docs.

PR #65, opened 2016-07-20
Update configuration.rb to be more readable

fidelisrafael:
Hi there! First of all, thanks for this amazing project. I'm looking at the source code, and I think that this file need a small update to be more readable! :+1:


danmayer:
awesome thanks

PR #66, opened 2016-07-20
Minor code improvement

fidelisrafael:
Just to keep more readable and clean.

### DIFF
--- a/lib/coverband/memory_cache_store.rb
+++ b/lib/coverband/memory_cache_store.rb
@@ -32,12 +32,7 @@ module Coverband
         #first time we see a file, we pre-init the in memory cache to whatever is in store(redis)
         line_cache = files_cache[file] ||= Set.new(store.covered_lines_for_file(file))
         lines.reject! do |line|
-          if line_cache.include? line
-            true
-          else
-            line_cache << line
-            false
-          end
+          line_cache.include?(line) ? true : (line_cache << line and false)
         end
         filtered_file_hash[file] = lines if lines.any?
       end


lib/coverband/memory_cache_store.rb
```diff
@@ -32,12 +32,7 @@ def filter(files)
         #first time we see a file, we pre-init the in memory cache to whatever is in store(redis)
         line_cache = files_cache[file] ||= Set.new(store.covered_lines_for_file(file))
         lines.reject! do |line|
-          if line_cache.include? line
-            true
-          else
-            line_cache << line
-            false
-          end
+          line_cache.include?(line) ? true : (line_cache << line and false)
```
lucasgomide: Maybe
```
return true if line_cache.include?(line)

line_cache << line
false
```

fidelisrafael: This is still weird, the same thing can be achived as single name: `line_cache.include?(line) ? true : (line_cache << line and false)` 
`line_cache#<<` will always return the new array value so `and` expression will be always executed, maybe this can be even better with:  
`line_cache.include?(line) ? true : (!!(line_cache << line) and false)`

fidelisrafael: And one more thing, inside this context(closure) using `return` keyword will throw a `(LocalJumpError)` exception.

lucasgomide: hm, sorry.. I saw only your changes.. I not saw the context. But this `line_cache.include?(line) ? true : (!!(line_cache << line) and false)` is not more readable than this 
```
if line_cache.include? line
  true
else
 line_cache << line
 false
end
```

fidelisrafael: I agree, that's why I prefer don't using in this way, but for me 

`line_cache.include?(line) ? true : (line_cache << line and false)` 

is very easy readable.

lucasgomide: Maybe move for another method could be more readable

lucasgomide: Yeap, i like your last sample

fidelisrafael: I don't believe this is the case, I mean, method call inside blocks normally is **very expensive**, this "sacrifice" can be acceptable in cases where the domain logic is too heavy or complex, see this benchmarks:

```ruby
require 'benchmark'

arry = (0..100_000).to_a
interations = (ENV['ITERATIONS'] || 1000).to_i

def reject_filter(value)
  return value > 50_000
end

Benchmark.bm(50) do |bm|
  bm.report('reject with method call inside block') do
    interations.times do |i|
      arry.reject do |value|
        reject_filter(value)
      end
    end
  end

  bm.report('reject without method call inside block') do
    interations.times do |i|
      arry.reject do |value|
        value > 50_000
      end
    end
  end

  bm.report('reject with method to proc call') do
    interations.times {
      arry.reject(&method(:reject_filter))
    }
  end
end
```

My results:

```
➔ ITERATIONS=1000 ruby test.rb
reject with method call inside block                 8.720000   0.030000   8.750000 (  8.754245)
reject without method call inside block              5.760000   0.040000   5.800000 (  5.801007)
reject with method to proc call                     14.930000   0.000000  14.930000 ( 14.928754)
```

lucasgomide: Yeap, I agree with you.. But my point was about the readability of code, and not about performance it. 
Everything know, than sometimes a ugly code is a faster code. hehe

But nice, good benchmark

fidelisrafael: Yes, just commented to better explain my motivations, and THIS:

> Everything know, than sometimes a ugly code is a faster code. hehe :clap:

danmayer:
ha nice discussion in this case I think since it is just used in test readability is the primary goal but happy with the change... so Thanks!

PR #69, opened 2016-10-17
Fix big data issue.

butcher:
Because we not collect line usage statistics, we can store evaluates lines only once.
I faced with big data issue when couple of iterators over 100 000 records kill the script.
As a result I can not write data into Redis (connection die with ECONNRESET).

So here is the fix and some test.


danmayer:
Thanks this makes sense sorry for the delay, have been a bit busy lately... I am planning to get back into some coverband development.

PR #70, opened 2016-11-04
Remove require json and inline parse_baseline

guigs:
json is already a runtime dependency


danmayer:
thanks for the cleanup

PR #75, opened 2016-12-02
Performance & memory usage improvements

guigs:
This is a combination of #72, #73 and #74, since they touch pretty much the same code, I've merged them here.

### DIFF
--- a/lib/coverband/base.rb
+++ b/lib/coverband/base.rb
@@ -1,4 +1,5 @@
 require 'singleton'
+require 'set'
 
 module Coverband
   class Base
@@ -41,6 +42,7 @@ module Coverband
       @files = {}
       @file_usage = Hash.new(0)
       @file_line_usage = {}
+      @ignored_files = Set.new
       @startup_delay = Coverband.configuration.startup_delay
       @ignore_patterns = Coverband.configuration.ignore + ["internal:prelude"]
       @ignore_patterns += ['gems'] unless Coverband.configuration.include_gems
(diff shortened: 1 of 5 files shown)


danmayer:
Awesome thanks. I will review tonight or tomorrow.

danmayer:
Great work!

Previous benchmark from `Rake Benchmark`

```
coverband         0.990000   0.010000   1.000000 (  1.066530)
no coverband      0.180000   0.000000   0.180000 (  0.187218)
```

with all your changes and the new feature (full counts)

```
coverband         0.810000   0.000000   0.810000 (  0.849571)
no coverband      0.190000   0.000000   0.190000 (  0.185481)
```

I bumped the version and released, moving to 1.3.0 as this changes data format and requires folks to clear their redis cache.

Pushed coverband 1.3.0 to rubygems.org.

guigs:
Awesome, thanks!

I didn't change data format in redis, because I keep passing only the array of touched lines, without counts (see https://github.com/danmayer/coverband/commit/390cce8c7a0d86cbf9bfe79b90fab6cbd5c52b65#diff-da719a3229eb818f0abe82bc2332abd8R13).

So in order to actually have line counts we still need to change how data is persisted to redis. We need to read current line counts from redis sum with current report values and store updated counts.

danmayer:
ahh makes sense missed that... oh well I guess I will make another version bump when we add that. Might actually be good to change base namespace as it would still be easy to have someone upgrade without noticing. Thanks, if you want to work on that cool, if not I will try to fit it in sometime soon... likely can't get to it until the weekend.

danmayer:
^^ @guigs

PR #79, opened 2016-12-28
initial work on pushing hit counts and retrieving them from redis

danmayer:
This works, but has made it obvious that we need to refactor the reporter class.

* It shouldn't be accessing Redis directly and should instead initialize a storage class. 
* We could then have abstract reporter, simplecov reporter, console reporter
* The reporter class shouldn't be doing the baseline, refactoring that to a baseline class which can use store would allow it to store in file, redis, or memory store
* reporter class should have some instance variables and move away from being all `self` methods
* currently tests around the reporter are very difficult to write / maintain

screenshot 2016-12-28 11 54 11


danmayer:
OK this has been a bigger refactoring and project than initially planned, getting pretty close to the next version worth releasing... I need to do some performance work on how we are storing the hit counts in redis as this initial version lost pipelining and significantly slowed down as a result.

danmayer:
OK after adding some benchmarks, and other tests this may be good to go... Thinking I will push this to one of my projects so I can see impacts... perhaps release as a release candidate first if anyone is interested in helping test it out.

danmayer:
testing on live services and will release shortly any comments appreciated!

danmayer:
merged and release, but if you have feedback please let me know... This is the start of some more significant changes.

PR #85, opened 2017-03-16
Coverband::Base#sample returns the result of the yield

kamal:
This PR makes `Coverband::Base#sample` return the result of the `yield`. This was especially useful for me when writing a Coverband Sidekiq client middleware, which is as simple as this:

```ruby
class CoverbandSidekiqMiddleware
  def call(*)
    Coverband::Base.instance.sample { yield }
  end
end
```

 According to Sidekiq's design, `call` must return the result of the `yield`.

### DIFF
--- a/lib/coverband/base.rb
+++ b/lib/coverband/base.rb
@@ -21,8 +21,9 @@ module Coverband
     def sample
       configure_sampling
       record_coverage
-      yield
+      result = yield
       report_coverage
+      result
     end
 
     def save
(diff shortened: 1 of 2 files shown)


danmayer:
Awesome, this looks great let me review along with some other recent requests and get this in a release soon.

Sorry for the delayed response and handling this merge request, we just had [our first baby](https://millermayersadventures.com/2017/02/23/theodore-harley-mayer/) so I am a bit slow to do any open source work.

danmayer:
sorry for the delay @kamal  this has been released in version 1.5.1

kamal:
Awesome. Thanks! Hope you're catching up on sleep :)

Issue #89, opened 2017-07-07
`rake coverbands:baseline` dropped production database.


stephenprater:
First:  Let me be clear - this is not actually the gems fault.  The gem has a relatively minor security "flaw" which resulted in an outsize result in our environment because of poor practices / bad luck on _our_ end.

In the `baseline` task, Coverband loads (and because it's Ruby, executes) every `.rb` file under both `/lib` and `/app` - including potentially destructive scripts that are not loaded by the regular load process.

In our case, we had `lib/<submodule/db/schema.rb` files from an old submodule that were still present in Capistrano's cached copy.  Although we had run Coveralls on staging environments to verify it's operation, since that file was only present on a single production server, when we executed the `baseline` task on that server, it was loaded and forced schema creation, dropping existing tables.

Although this is definitely one of those "swiss cheese holes align" kind of problems - I think Coverband could benefit from being more intentional about what files it executes while establishing the baseline coverage.

Which of the following approaches would be most compatible with Coverband's overall philosophy?

1. Do not descend into submodule directories, and do not execute Ruby found there.
2. Maintain a 'blacklist' of "dangerous" files that should never be executed.
3. Open the database in "read only" mode while establishing a baseline.
4. Hook into the existing Rails eager load - which should eager load all files _referenced_ by the code base without loading potentially dangerous 'sidecar' scripts.

latazzajones:
We've just been debugging a very similar issue - fortunately we caught ours in development. BUT! This seems like something that should be rectified or documented. In our case, we have an engine in our lib file with a dummy app and that schema is run by coverband. 

A blacklist seems like a good approach to this issue but this is also super dangerous behavior.

danmayer:
WOW, prod DB drop that is terrible. So sorry about that... I am a bit confused what has changed to have two folks hit this recently, as I have never had a report like this in the 4 years the project has been in use.

I will need to ponder a bit... Some quick thoughts:

* I run baseline on dev machine and not on production, I just send it to the same source as when I report from production
    * If that isn't clear from the docs it might be good to make that more clear, there isn't any reason to run `baseline` on production (or staging, QA stacks, etc) EVER.
* I guess most ruby should 'execute' on load, but generally it shouldn't have any actions. I definitely consider any scripts that do pretty dangerous
* I don't like the blacklist as it would mean you need to know what to avoid before running. So I think we will need to come up with something better and safer than that.
   * That being said since both issues are schema related perhaps skipping all schema files by default would be a smart move.
[...]

PR #90, opened 2017-07-10
Do not require all files if eager load is on

stephenprater:
As detailed in #89 - This changes coverband to not seek out and require files which are not loaded by the Rails loading facility.

### DIFF
--- a/lib/coverband/tasks.rb
+++ b/lib/coverband/tasks.rb
@@ -20,20 +20,23 @@ namespace :coverband do
             require baseline_file
           end
         end
-        if defined? Rails
-          Dir.glob("#{Rails.root}/app/**/*.rb").sort.each { |file| 
+        if defined? Rails && !Rails.config.eager_load
+          Dir.glob("#{Rails.root}/app/**/*.rb").sort.each do |file|
               begin
                 require_dependency file
               rescue Exception
                 #ignore
-              end }
+              end
+          end
+
           if File.exists?("#{Rails.root}/lib")
-            Dir.glob("#{Rails.root}/lib/**/*.rb").sort.each { |file|
+            Dir.glob("#{Rails.root}/lib/**/*.rb").sort.each do |file|
               begin
                 require_dependency file
               rescue Exception
                 #ignoring file
-              end}
+              end
+            end
           end
         end
       }


stephenprater:
Not sure what's happening with CI - I did not touch that part.

lib/coverband/tasks.rb
```diff
@@ -20,20 +20,23 @@
             require baseline_file
           end
         end
-        if defined? Rails
-          Dir.glob("#{Rails.root}/app/**/*.rb").sort.each { |file| 
+        if defined? Rails && !Rails.config.eager_load
```
danmayer: Does this mean when `Rails.config.eager_load` is false coverband wouldn't load anything? I guess basically in that case baseline just doesn't work...

lib/coverband/tasks.rb
```diff
@@ -20,20 +20,23 @@
...
+          Dir.glob("#{Rails.root}/app/**/*.rb").sort.each do |file|
               begin
                 require_dependency file
               rescue Exception
                 #ignore
-              end }
+              end
+          end
+
           if File.exists?("#{Rails.root}/lib")
```
danmayer: Should this `&& !Rails.config.eager_load` also be added here

danmayer:
added some questions let me know your thoughts @stephenprater

lib/coverband/tasks.rb
stephenprater: No - this means that if Rails is defined and `eager_load` is TRUE then Coverband wouldn't load anything itself (since it would all have been required by the rails boot process.)

`eager_load` is generally true in prod-like and test environments.

lib/coverband/tasks.rb
stephenprater: nah - both of these things are skipped if eager load is true or if we're not in Rails.  some people (yours truly) stupidly keep destructive files not loaded in the normal course of things in the lib directory.  ;)

stephenprater:
@danmayer I think the other PR is sufficient - perhaps with a documentation note to not run baseline in prod - but to point to the same location as when you run locally.

danmayer:
ok cool thanks

PR #91, opened 2017-07-12
Exclude files in config ignore for baseline

latazzajones:
👋 This pull request updates the baseline task to exclude files that are excluded by Coverband.configuration.ignore. 

The idea here is two-fold: 
1. Why do we need to include these files in the baseline if they're being ignored in the coverage reporting?

2. Files that get executed on initialization can be added here. This includes stuff like schema's for dummy apps in engines and other executable scripts. Thus - preventing the baseline task from destroying a person's database. 

I'm not sure that this is the final solution to https://github.com/danmayer/coverband/issues/89 - ultimately we should be blocking anything from being executed. But - this is a stopgap and hopefully also seems like a logical addition to the task.

### DIFF
--- a/lib/coverband/baseline.rb
+++ b/lib/coverband/baseline.rb
@@ -4,7 +4,6 @@ module Coverband
     def self.record
       require 'coverage'
       Coverage.start
-      #binding.pry
       yield
 
       project_directory = File.expand_path(Coverband.configuration.root)
--- a/lib/coverband/baseline.rb
+++ b/lib/coverband/baseline.rb
@@ -4,6 +4,7 @@ module Coverband
     def self.record
       require 'coverage'
       Coverage.start
+      #binding.pry
       yield
 
       project_directory = File.expand_path(Coverband.configuration.root)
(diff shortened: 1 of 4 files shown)


lib/coverband/baseline.rb
```diff
@@ -4,6 +4,7 @@ class Baseline
     def self.record
       require 'coverage'
       Coverage.start
+      #binding.pry
```
danmayer: remove

lib/coverband/tasks.rb
```diff
@@ -5,6 +5,7 @@
     Coverband::Baseline.record {
         if Rake::Task.tasks.any?{ |key| key.to_s.match(/environment$/) }
           Coverband.configuration.logger.info "invoking rake environment"
+          #breaks our app here
```
danmayer: remove

latazzajones:
@danmayer Comment removed! Thanks for catching that :-)

danmayer:
sorry @latazzajones one more comment to remove I added in the diff.

Thanks so much for this PR

lib/coverband/baseline.rb
danmayer: this is the remaining one

latazzajones: 🤦‍♀️ Done!

danmayer:
coverband 1.5.2 released with this feature

stephenprater:
@latazzajones Thanks for this PR.  🍰  for you - or the dietarily appropriate item of your choice.

PR #98, opened 2018-02-27
Able to set custom AWS S3 settings for coverband

a0s:
Add support for separate AWS S3 settings of region/bucket/key/secret for coverband in case when we don't want to use default AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY. 

Old behaviour, reading from AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY:
```ruby
  ...
  config.s3_bucket = 'bucket_name'
  ...
end
```

New behaviour, set all settings for coverband only:
```ruby
  ...
  config.s3_bucket = ENV['COVERBAND_AWS_BUCKET']
  config.s3_region = ENV['COVERBAND_AWS_REGION']
  config.s3_access_key_id = ENV['COVERBAND_AWS_KEY']
  config.s3_secret_access_key = ENV['COVERBAND_AWS_SECRET']
  ...
end
```

### DIFF
--- a/lib/coverband/configuration.rb
+++ b/lib/coverband/configuration.rb
@@ -4,8 +4,8 @@ module Coverband
     attr_accessor :redis, :root_paths, :root,
                   :ignore, :additional_files, :percentage, :verbose, :reporter,
                   :stats, :logger, :startup_delay, :trace_point_events,
-                  :include_gems, :memory_caching, :s3_bucket, :coverage_file, :store,
-                  :disable_on_failure_for
+                  :include_gems, :memory_caching, :coverage_file, :store, :disable_on_failure_for,
+                  :s3_bucket, :s3_region, :s3_access_key_id, :s3_secret_access_key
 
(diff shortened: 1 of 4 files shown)


danmayer:
I will address this in a quick follow up after getting 2.0 released...

PR for Coverband 2.0! Alpha release running on prod for last 2 months. Let me know if there is any last minute feedback, if not I will release later tonight ;) Also, sharing thoughts Future 3.0 roadmap. 

a0s:
For now, we are using 1.5.4 with this PR. So it would be great to include this into 2.0 ...

danmayer:
OK, this is merged into the 2.0 line, on the master branch. I am going to handle some of the other PRs and get a 2.x.x out in a few days, which will include this work. Thanks @a0s

danmayer:
this is in the 2.0.1 release https://github.com/danmayer/coverband/blob/master/changes.md#201

Issue #99, opened 2018-03-05
Unable to start coverband from outside the Rails project directory


ThomasOwens:
I noticed a few things, all in lib/coverband.rb.

First, CONFIG_FILE is hard coded to a relative path. This is all well and good until you try to invoke a Rails application from outside Rails.root. From what I can tell, Coverband is looking for the CONFIG_FILE relative to the location where you run your script. Although I can set my working directory before starting the application, this doesn't seem like it should be necessary.

I also noticed that the self.configure method takes in a file argument, but it is unused. It appears like it should allow a caller to, instead of passing in a block, pass in a path to a configuration file instead. This would add a great deal of flexibility into users of Coverband by allowing them to choose a location to put the configuration file, instead of requiring it to be in './config/coverband.rb'.

ThomasOwens:
I've opened one possible solution as a PR: https://github.com/danmayer/coverband/pull/100

danmayer:
excellent, I will address this and help get that merged in after I get the long in progress Coverband 2.0 PR released. Thanks so much for the interest and support.

https://github.com/danmayer/coverband/pull/107

danmayer:
resolved closing this out

PR #100, opened 2018-03-05
#99 Using the file argument to self.configure in lib/coverband.rb

ThomasOwens:
This allows a caller of Coverband.configure to not only pass in a block or have a file in ./config/coverband.rb, but to put the configuration file anywhere they want and pass it in as an argument. If the argument is unused, then the existing behavior is preserved. This should make it easier to invoke an application that uses Coverband from any working directory.

### DIFF
--- a/lib/coverband.rb
+++ b/lib/coverband.rb
@@ -38,7 +38,7 @@ module Coverband
       if File.exists?(configuration_file)
         require configuration_file
       else
-        raise ArgumentError, "configure requires a block or the existance of a #{CONFIG_FILE} in your project"
+        raise ArgumentError, "configure requires a block, the existance of a #{CONFIG_FILE} in your project, or a path to a config file passed in to configure"
       end
     end
   end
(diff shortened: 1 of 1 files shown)


danmayer:
Thanks @ThomasOwens  I have been behind on taking PRs are my other branch makes some major breaking changes, but I will incorporate this soon. The Alpha branch is just about ready to go, I updated the readme to match the new version... https://github.com/danmayer/coverband/tree/feature/via_coverage#installation

danmayer:
OK @ThomasOwens a PR for Coverband 2.0! Alpha release running on prod for last 2 months. Let me know if there is any last minute feedback, if not I will release later tonight ;) Also, sharing thoughts Future 3.0 roadmap. 

after that I will come back to this PR

danmayer:
OK, this is merged into the 2.0 line, on the master branch. I am going to handle some of the other PRs and get a 2.x.x out in a few days, which will include this work. Thanks  @ThomasOwens

danmayer:
this is in the 2.0.1 release https://github.com/danmayer/coverband/blob/master/changes.md#201

PR #102, opened 2018-07-02
Add option to periodically flush memory to persistent storage, instea…

oded-zahavi:
…d of flushing it for each change in state

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,4 +2,3 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in coverband.gemspec
 gemspec
-
--- a/coverband.gemspec
+++ b/coverband.gemspec
@@ -27,8 +27,9 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'sinatra'
   spec.add_development_dependency 'classifier-reborn'
   spec.add_development_dependency 'aws-sdk', '~> 2'
-  spec.add_runtime_dependency "simplecov", '> 0.11.1'
-  spec.add_runtime_dependency "json"
+  spec.add_runtime_dependency     'simplecov', '> 0.11.1'
+  spec.add_runtime_dependency     'activesupport', '> 4.1'
+  spec.add_runtime_dependency     'json'
   # TODO make redis optional dependancy as we add additional adapters
-  spec.add_runtime_dependency "redis"
+  spec.add_runtime_dependency     'redis'
 end
--- a/lib/coverband/adapters/memory_cache_store.rb
+++ b/lib/coverband/adapters/memory_cache_store.rb
@@ -1,10 +1,15 @@
+require 'active_support'
+
 module Coverband
   module Adapters
     class MemoryCacheStore
       attr_accessor :store
 
-      def initialize(store)
-        @store = store
+      def initialize(store, options = {})
+        @store       = store
+        @max_caching = options[:max_caching] || 1
+        @count       = 0
+        @filtered_files = {}
       end
 
       def self.reset!
@@ -16,8 +21,13 @@ module Coverband
       end
 
       def save_report(files)
-        filtered_files = filter(files)
-        store.save_report(filtered_files) if filtered_files.any?
+        @count = @count.succ % @max_caching
+        @filtered_files.deep_merge!(filter(files))
+
+        return unless @count.zero?
+        return unless @filtered_files.any?
+        store.save_report(@filtered_files)
+        @filtered_files.clear
       end
 
       private
--- a/lib/coverband/base.rb
+++ b/lib/coverband/base.rb
@@ -48,7 +48,7 @@ module Coverband
       @ignore_patterns += ['gems'] unless Coverband.configuration.include_gems
       @sample_percentage = Coverband.configuration.percentage
       @store = Coverband.configuration.store
-      @store = Coverband::Adapters::MemoryCacheStore.new(@store) if Coverband.configuration.memory_caching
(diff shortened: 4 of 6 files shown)


danmayer:
I will take a closer look tomorrow, but you probably want to just wait or perhaps help me test the alpha branch, it reworks much of coverband and is far more performant than the current release version... check out the https://github.com/danmayer/coverband/tree/feature/via_coverage branch... I haven't gotten to updating all the documentation so it might be a bit hard to figure out... I can try to finish that up but probably not for a week as I am traveling for work.

oded-zahavi:
Thanks! I'd love to help out and contribute!

oded-zahavi:
Also if you are opened for that I'd also suggest integrating message broker adapter such as RabbitMQ or Kafka to make persisting totally asynchronous - wdyt?

danmayer:
Thanks @oded-zahavi   I have been behind on taking PRs are my other branch makes some major breaking changes, but I will incorporate this soon. The Alpha branch is just about ready to go, I updated the readme to match the new version... https://github.com/danmayer/coverband/tree/feature/via_coverage#installation

Once I get this release out I would be very interested in looking at other adapters like RabbitMQ or Kafka

danmayer:
OK @oded-zahavi A PR for Coverband 2.0! Alpha release running on prod for last 2 months. Let me know if there is any last minute feedback, if not I will release later tonight ;) Also, sharing thoughts Future 3.0 roadmap. 

after I get that up I will get back to the blocked PRs, and I would be happy to pair up a bit to show how the new data adapters should make it pretty easy to add Kafka or RabbitMQ storage.

oded-zahavi:
Hey thanks I'll need a day or two to learn the new code. Hope to come up with some thoughts after :)

danmayer:
sounds good @oded-zahavi no rush

oded-zahavi:
Hey so I think there's a misunderstanding about what I did: I am aware of `sampling_rate`. The thing is that even when sampling 1% or even less, performance wise it would be very beneficial to send aggregated info to redis in one bulk. The way I see it, realtime info is less important to coverband users. Performance and effect on response time, however, are. Our #1 concern when using coverband was performance and whether it would affect our production systems. I will have to rebase my pull request on top of master, but before moving on I'd like to hear your thoughts :)

danmayer:
yeah agreed that folks don't need realtime data for Coverband, generally just looking at snapshots of the data. I think there is still some confusion, basically `sampling_rate` was the old coverband 1.x concept. The same value in the configuration file now is really what you implemented here aggregated data send frequency. If you are using Coverband 2.0 with the `Coverage` collector, it does exactly what this PR implements, IE it reduces real-time production overhead by only sending data in large batches of in-memory data at the frequency configured. For example on production my data send frequency is now 0.01, which means I send data every 1000 requests or so. 

Does that make sense?

danmayer:
thoughts on getting this in for a 2.0.3 release @kbaum / @oded-zahavi since you all are already running it on a branch and I will be making at least one more release before 3.0 is ready... Could you all fix it up to get a clean merge and get tests passing again?

kbaum:
Honestly I am not confident in the memory cache store on master because when I made my last set of changes to it, I didn’t realize that it was receiving deltas in coverage each time as opposed to all of the coverage data. To be honest, I think we shouldn’t need memory cach store if we are always reporting to redis periodically in a background thread.

danmayer:
OK thanks @kbaum I will ignore this and the other memory cache PR while trying to get 3.0 out, I am going to see if there is anything else I could do quickly to improve perf for the 2.0.3 release

kbaum:
Re: performance improvements. Does the current coverband send file coverage data to redis when there’s been no coverage changes for that file? In other words does it just send 0’s to redis?

danmayer:
yeah @kbaum that is where I still think the memory cache could help with 2.0.3 as it sends a lot of generic ruby files that aren't hit in the avg request... but that being said I tried to make a quick perf improvement and realized how much easier it will be after simplifying, so am trying to decide how much more effort to put into that or just push harder for 3.0

danmayer:
Yeah I think that might work well... If we can use memory cache to do it...
I don't know which parts and pieces to pull off memory cache to get that
going, I was trying something else. If you know where we stand between what
is on master and your two memory related PRs and can get tests passing it
would be good to try. I could update the benchmark to change 10% of the
files or something every time it reports.

danmayer:
OK I pushed a new benchmark that should help us verify memory cache improvements... See this commit https://github.com/danmayer/coverband/commit/a6cb986194c879f7a5c863f57bc1de811db3a3ac

This shows even the current version of the memory store on master is quite effective (7x faster when 15% of the files are changing), I wonder if for version 2.x.x, just defaulting the memory cache on would perhaps help most folks enough.

```
rake benchmarks:cache_reporting
runs benchmarks on reporting large files to redis using memory cache
 store_redis_reports     1.000  i/100ms
 store_cache_reports     1.000  i/100ms
 store_redis_reports      0.748  (± 0.0%) i/s -     12.000  in  16.129973s
 store_cache_reports      5.777  (±17.3%) i/s -     81.000  in  15.154805s

Comparison:
 store_cache_reports:        5.8 i/s
 store_redis_reports:        0.7 i/s - 7.72x  slower
```

I believe the benchmark is sowing real impacts as adjusting the percent fake change to 100% and all the perf win disappears.

danmayer:
I think this is basically covered by the new background thread so I am closing this we can revisit if there is any reason to.

oded-zahavi:
Cool :)

Issue #97, opened 2017-10-18
Running baseline in production


kbaum:
The coverband README warns against running `rake coverband:baseline` in production. I would like to schedule something like this to run every couple of weeks:

```
rake coverband:clear 
rake coverband:baseline 
```
Whats the danger of running `coverband:baseline` in production?

danmayer:
So basically it comes back to this ticket, where the folks had a checked in file that when loaded would wipe the DB.... o_0

https://github.com/danmayer/coverband/issues/89

Coverband baseline tries to load all files we added some features to prevent that, but basically, you can run the baseline in dev mode (or in a staging environment) or whatever else to get the same results and then still merge it with the runtime data. If you know you don't have any dangerous files you can run it in production or you could try just running it like so `RAILS_ENV=dev rake coverband:baseline`

danmayer:
let me know if that helps or if I should try to make it more clear in the readme

lacco:
Just to clarify my understanding about the baseline: Although the README states that you should not run the rake task on production, we need to have a way to record the baseline.

Running `RAILS_ENV=development rake coverband:baseline` is no option for me because this environment is not set up there. Running `rake coverband:baseline` locally and just passing the production redis options is also not working since our production redis is protected from outside access.

As far as I see the only option for me is to actually run `RAILS_ENV=production rake coverband:baseline` which should not be an issue as long as it's safe to load all files from `/app` and `/lib`?

danmayer:
yes, this is confusing for now especially with the baseline now also stored in Redis it used to be much easier to merge a dev coverage baseline and production coverage data. Long as you don't have any dangerous files it should be safe to run. The next version of Coverband that I am working on will no longer require a baseline at all which should simply all of these issues. It will still take me awhile to get that version out though as I am only working on it a bit here and there.

ndbroadbent:
Reading through [the rake task code](https://github.com/danmayer/coverband/blob/master/lib/coverband/tasks.rb#L20), I can see why that would happen, since it was requiring `/app/**/*.rb` and `./lib/**/*.rb`.

I overrode the rake task in my app and replaced those lines with `Rails.application.eager_load!`. That seems to be working quite well for me. There are some files in `app` and `lib` that weren't eager loaded, but if they're still 0% after some web requests then it would be easier to see that they aren't being used.

danmayer:
[...]

Issue #88, opened 2017-06-29
Incorrectly reporting 100% coverage


jagthedrummer:
We're running into an issue where coverband reports data incorrectly. Here's a screenshot that illustrates the problem.

<img width="1353" alt="screen shot 2017-06-29 at 10 36 31 am" src="https://user-images.githubusercontent.com/58702/27697383-af77ffd4-5cb9-11e7-9317-c3ae1827af3c.png">

Coverband says that the class is 100% covered, but I _know_ that I didn't make enough calls to exercise the entire controller, and there are many lines that indicate they've never been run, but that don't count as being "missed" for some reason.

This particular controller lives outside of `/app/` in a directory that is treated as an "internal gem", so to speak. That directory is pulled into the `Gemfile` like this:

```ruby
gemspec path: 'userpages'
```

To get any coverage data at all for anything in that directory I added this to my coverband config.

```ruby
config.include_gems = true
config.root_paths = ['/app/','/userpages/']
```

danmayer:
odd it looks like it is considering those lines as comments / non active lines like whitespace. I will have to take a closer look to see what is going on. Thanks for the report, guessing it is related to the gem coverage, I will see if I can recreate.. Do you know of any opensource gem that might be close to the internal gem you use... to help me recreate @jagthedrummer

jagthedrummer:
@danmayer I don't know of any open source gems that would be close to what we're doing. I'm not sure if any gem that provides a controller would do it, or if there's something else peculiar about our gem.

danmayer:
OK I will see if I can figure out why this would happen..... not sure I will be able to debug without a way to reproduce.

jagthedrummer:
@danmayer I've created a sample repo that demonstrates this problem.

https://github.com/jagthedrummer/coverband-internal-gem-demo

The project contains a directory called `interal_full_mountable_gem` that provides a controller and some routes for the main app. (I misspelled 'internal' and didn't catch it until I was all done. Oops.)

The internal gem is loaded by this line in `Gemfile`:

```ruby
gemspec path: 'interal_full_mountable_gem'
```

The routes are hooked up with this line in `config/routes.rb`:

```ruby
mount InteralFullMountableGem::Engine, at: '/'
```

After installing and configuring everything I ran: `rake coverband:baseline`

Then started the rails server and visited: `http://localhost:3000/gem_test/test1`

Then `rake coverband:coverage`

I checked in the `coverage/` directory that I generated from coverband.

[...]

PR #105, opened 2018-07-22
Use Rails.application.eager_load! instead of manually requiring app/**/*.rb and lib/**/*.rb

ndbroadbent:
See: #97

I think it might be a better idea to use `Rails.application.eager_load!`, which only loads code that your application is actually using. This is safer than loading all Ruby files in `app/**/*.rb` or `lib/**/*.rb`, because these folders might contain some Ruby scripts that could have unintended consequences.

### DIFF
--- a/lib/coverband/tasks.rb
+++ b/lib/coverband/tasks.rb
@@ -39,12 +39,7 @@ namespace :coverband do
 
       safely_import_files(Coverband.configuration.additional_files.flatten)
 
-      if defined? Rails
-        safely_import_files(Dir.glob("#{Rails.root}/app/**/*.rb"))
-        if File.exists?("#{Rails.root}/lib")
-          safely_import_files(Dir.glob("#{Rails.root}/lib/**/*.rb"))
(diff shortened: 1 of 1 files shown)


ndbroadbent:
Ohhhhh, I also started running into #88 and now I understand that all the files must be loaded in the baseline, otherwise it just ignores all of those files and marks most of the code as "irrelevant". Not really sure why that's happening, but I guess that makes sense!

danmayer:
Let me take a closer look @ndbroadbent at this work and the related issue after getting 2.0 out...

https://github.com/danmayer/coverband/issues/97

Basically with the way 2.0 works, I think I can entirely remove the concept of the baseline recording and introduce a new concept for any file that were not getting picked up do to load order issues. Especially with the new `Coverage` support, the load order of files related to starting Coverband is extremely important... Although I think allowing devs to config a force_file_reload list with directory globs might be a good compromise of always use `eager_load` and then be explicit when something is missed.

PR #109, opened 2018-08-08
Fix type error

viktor-silakov:
Prevent from `TypeError: nil can't be coerced into Integer` if  `original` array has `nil` value at this index.

### DIFF
--- a/lib/coverband/collectors/coverage.rb
+++ b/lib/coverband/collectors/coverage.rb
@@ -58,7 +58,7 @@ module Coverband
       private
 
       def array_diff(latest, original)
-        latest.map.with_index { |v, i| v ? v - original[i] : nil }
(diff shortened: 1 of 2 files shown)


danmayer:
I have never hit this error on any projects but it makes sense did you hit this while upgrading or do you know more about the cause? @viktor-silakov 

same, on this looks good and I will get it into the 2.0.2 release

viktor-silakov:
Yes, i got this after upgrade to 2.0.0. But i didn't do a deep investigate.

danmayer:
ahh OK, so I think I didn't make it clear enough, that when upgrading from 1.X to 2.x one needs to run `rake coverband:clear` as the data isn't entirely compatible. Your fix will prevent errors but might give non-accurate results, until it is cleared ;) 

Thanks @viktor-silakov 

Still best to avoid the error I think, but I can try to make it more clear about changing data on upgrade.

PR #110, opened 2018-08-09
configure logs levels, remove unnecessary log messages

viktor-silakov:
Log is too verbose. I got more that 5 GB this night. Configured some message levels.

### DIFF
--- a/lib/coverband/collectors/base.rb
+++ b/lib/coverband/collectors/base.rb
@@ -87,7 +87,7 @@ module Coverband
       end
 
       def output_file_line_usage
-        @logger.info 'coverband debug coverband file:line usage:'
+        @logger.debug 'coverband debug coverband file:line usage:'
         @file_line_usage.sort_by { |_key, value| value.length }.each do |pair|
           file = pair.first
           lines = pair.last
--- a/lib/coverband/collectors/coverage.rb
+++ b/lib/coverband/collectors/coverage.rb
@@ -18,7 +18,7 @@ module Coverband
         end
 
         if failed_recently?
-          @logger.info 'coverage reporting standing-by because of recent failure' if @verbose
+          @logger.error 'coverage reporting standing-by because of recent failure' if @verbose
           return
         end
 
@@ -39,8 +39,8 @@ module Coverband
           @store.save_report(@file_line_usage)
           @file_line_usage.clear
         elsif @verbose
-          @logger.info 'coverage report: '
-          @logger.info @file_line_usage.inspect
+          @logger.debug 'coverage report: '
+          @logger.debug @file_line_usage.inspect
         end
       # StandardError might be better option
       # coverband previously had RuntimeError here
@@ -49,9 +49,9 @@ module Coverband
       rescue StandardError => err
         failed!
         if @verbose
-          @logger.info 'coverage missing'
-          @logger.info "error: #{err.inspect} #{err.message}"
-          @logger.info err.backtrace
+          @logger.error 'coverage missing'
+          @logger.error "error: #{err.inspect} #{err.message}"
+          @logger.error err.backtrace
         end
       end
 
--- a/lib/coverband/collectors/trace.rb
+++ b/lib/coverband/collectors/trace.rb
@@ -21,9 +21,9 @@ module Coverband
       rescue RuntimeError => err
         failed!
         if @verbose
-          @logger.info 'error stating recording coverage'
-          @logger.info "error: #{err.inspect} #{err.message}"
-          @logger.info err.backtrace
+          @logger.error 'error stating recording coverage'
+          @logger.error "error: #{err.inspect} #{err.message}"
+          @logger.error err.backtrace
         end
       end
 
@@ -40,7 +40,7 @@ module Coverband
         unset_tracer
 
         if failed_recently?
-          @logger.info 'coverage reporting standing-by because of recent failure' if @verbose
+          @logger.error 'coverage reporting standing-by because of recent failure' if @verbose
           return
         end
 
@@ -59,9 +59,9 @@ module Coverband
       rescue RuntimeError => err
         failed!
         if @verbose
-          @logger.info 'coverage missing'
-          @logger.info "error: #{err.inspect} #{err.message}"
-          @logger.info err.backtrace
+          @logger.error 'coverage missing'
+          @logger.error "error: #{err.inspect} #{err.message}"
+          @logger.error err.backtrace
         end
       end
 
--- a/lib/coverband/reporters/base.rb
+++ b/lib/coverband/reporters/base.rb
@@ -9,13 +9,13 @@ module Coverband
 
         if Coverband.configuration.verbose
           Coverband.configuration.logger.info "fixing root: #{roots.join(', ')}"
-          Coverband.configuration.logger.info "additional data:\n #{additional_coverage_data}"
+          Coverband.configuration.logger.debug "additional data:\n #{additional_coverage_data}"
         end
 
         scov_style_report = report_scov_with_additional_data(store, additional_coverage_data, roots)
 
         if Coverband.configuration.verbose
-          Coverband.configuration.logger.info "report:\n #{scov_style_report.inspect}"
+          Coverband.configuration.logger.debug "report:\n #{scov_style_report.inspect}"
         end
         scov_style_report
       end
--- a/lib/coverband/tasks.rb
+++ b/lib/coverband/tasks.rb
@@ -9,8 +9,8 @@ namespace :coverband do
           require_dependency file
         rescue Exception => err
           if Coverband.configuration.verbose
-            Coverband.configuration.logger.info "error adding file to baseline: #{file}"
-            Coverband.configuration.logger.info "error: #{err}"
+            Coverband.configuration.logger.error "error adding file to baseline: #{file}"
+            Coverband.configuration.logger.error "error: #{err}"
           end
         end
       end
--- a/test/unit/collectors_base_test.rb
+++ b/test/unit/collectors_base_test.rb
@@ -63,7 +63,6 @@ class CollectorsBaseTest < Test::Unit::TestCase
     assert_equal false, coverband.instance_variable_get('@enabled')
     logger.expects(:info).at_least_once
     coverband.start
-    logger.info(1 + 1)
     coverband.stop
     coverband.save
   end
--- a/test/unit/collectors_coverage_test.rb
+++ b/test/unit/collectors_coverage_test.rb
@@ -75,7 +75,6 @@ if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.3.0')
       assert_equal false, coverband.instance_variable_get('@enabled')
       logger.expects(:info).at_least_once
       coverband.start
-      logger.info(1 + 1)
       coverband.stop
       coverband.save
     end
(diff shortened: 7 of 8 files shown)


danmayer:
gotcha log levels make sense @viktor-silakov  I don't recommend having verbose logging on in production, it is generally just used for debugging on a local dev setup and disabled on production.

Log levels could be nice and allow some level of production logging, but most of the time I imagine short of debugging one would want to leave them off as all the real data should end up in the final coverage reports.

Thoughts?

viktor-silakov:
@danmayer  i agree with you but... i don't use production environment, i use coverband to measure functional test (selenium, API etc) code coverage on test environment, this test can running over 10 - 20 hours and full debug log can be more that 5 - 10 Gb. I have no free space :) I configure config such way:
````
config.logger = ActiveSupport::Logger.new("#{Rails.root}/log/#{Rails.env}_coverage.log")
config.logger.level = Logger::INFO
```` 
This way i collect only WARN + logs. This is very useful commit for me.

**Comment by danmayer** [comment]
OK, gotcha makes sense I will try to review more closely tonight and get this merged for 2.0.2

I guess generally as I only use the Coverband logging during setup and configuration if you are using it in CI against test / staging server. Once you get it configured you still might want to disable logging. I turn it back on if there is a section of code I don't understand why it isn't getting recorded, but do you find other good value from the logs?

**Comment by viktor-silakov** [comment]
there is two thing:
1. debugging during development
2. troubleshooting errors, in my case errors raised a few hours later after the tests started, and i very needed in verbose error and stacktrace.

**Comment by danmayer** [comment]
OK sounds good I should get it merged this weekend, I can likely get a release of 2.0.2 out next week thanks for all the feedback.

**Comment by danmayer** [comment]
hmm I see changed to the file adapter, are you using the file adapter when running your tests? I would highly recommend moving to the Redis adapter. The file adapter is really for local dev or tests, but if you are running a staging server with multiple processes / threads the file will cause race conditions as you have seen and likely slow your server down a good deal. I am going to try to pull in your log level changes, but if you want to do more work on the file adapter it might be good to put that on a different branch. Let me know your thoughts  @viktor-silakov

**Comment by danmayer** [comment]
The first 2 commits are now merged on master related to logging improvements (9ced825, 2c68007) if you let me know what you are trying to do with the rest I might understand, but please try the Redis adapter

**Comment by viktor-silakov** [comment]
1. @danmayer  in my azure vm environment the redis adapter make over 3 seconds overhead within every http request. Why? I don't know, currently i investigate this. I will not plan rework file adapter, it make over 1 second overhead per http request. These were experimental commits.
2. in result [diff ](https://github.com/danmayer/coverband/pull/110/files)i don't see any mention about file adapter i removed them all in the next commits you can merge them all.

**Comment by danmayer** [comment]
hmmm @viktor-silakov that is interesting, I can't think of what is possibly going on that would cause 3 seconds of Redis activity... If using a reporting frequency it shouldn't even connect to redis on every request. If you could share your full `coverband/config.rb` here or emailing it to me might help me understand what all is going on.

**Comment by viktor-silakov** [comment]
```
  config.root = Dir.pwd
  config.collector = 'coverage'
  config.store = Redis.new()

  config.root_paths = ['/app/'] # /app/ is needed for heroku deployments
  config.ignore = ['lib/scrazy_i18n_patch_thats_hit_all_the_time.rb']
  config.startup_delay = Rails.env.production? ? 0 : 0

  config.percentage = Rails.env.production? ? 100.0 : 100.0

  config.logger = ActiveSupport::Logger.new("#{Rails.root}/log/#{Rails.env}_coverage.log")
  config.logger.level = Logger::INFO
  config.verbose = false
end
```

**Comment by danmayer** [comment]
OK, thanks, I will dig in more soon.

**Comment by viktor-silakov** [comment]
i think that hash is not better type for this case. 5000 hgetall is nearly 1.2 sec. I thing it be better use string type and json.

**Comment by viktor-silakov** [comment]
i write small benchmark that emulate maximally optimized redis store `save_report` flow (i use pipline and bulk operation cashing etc..):

```
require 'benchmark'
require 'redis'

class FakeRedis
  def mapped_hmset(key, values)
    sleep 0
  end

  def sadd(key, values)
    sleep 0
  end

  def hgetall(key)
    { "1" => 0, "3" => 0, "4" => 0, "5" => 0, "6" => 0, "7" => 0, "8" => 0, "9" => 0, "11" => 0, "14" => 0, "15" => 0, "18" => 0, "19" => 0, "22" => 0, "24" => 0, "25" => 0, "28" => 0, "29" => 0, "32" => 0, "33" => 0 }
  end

  def multi
    yield(self)
  end

  def config(command, key, val)

  end

end

class Hash
  def value
    return self
  end
end

# to show only ruby overhead
@redis ||= FakeRedis.new
#@redis ||= Redis.new

def redis
  @redis
end

def base_key
  @base_key ||= 'coverband2.'
end

def log(msg)
  puts msg
end

def existing_coverage(keys)
  redis.config(:set, 'hash-max-ziplist-value', '1024')
  out = {}

  redis.multi do |mult|
    keys.each do |key|
      out["#{base_key}.#{key}"] = mult.hgetall("#{base_key}.#{key}")
    end

  end
  out2 = {}
  out.keys.each { |key| out2[key] = out[key].value }
  out2
end

def save_report(report)
  log '@@@existing coverage'
  log Benchmark.measure {
    @existing_coverage = existing_coverage(report.keys)
  }
  log '@@@array'
  log Benchmark.measure {
    store_array(base_key, report.keys)
  }

  maps = []
  log '@@@each'
  log Benchmark.measure {
    report.each do |file, lines|
      maps << create_map("#{base_key}.#{file}", lines)
    end
  }

  log '@@@store'
  log Benchmark.measure {
    store_maps(maps)
  }
end

def create_map(key, values)
  unless values.empty?
    values = Hash[values.map { |k, val| [k.to_s, val] }]

    existing = @existing_coverage[key]
    values.merge!(existing) { |_k, old_v, new_v| old_v.to_i + new_v.to_i }
  end
  ({ key => values })
end

def store_maps(maps)
  redis.multi do |multi_redis|
    maps.each do |hash|
      key = hash.keys.first
      values = hash[key]
      next if values.empty?
      multi_redis.mapped_hmset(key, values)
    end
  end
end

def store_array(key, values)
  redis.sadd(key, values) unless values.empty?
  values
end


require 'json'
report = JSON.parse(File.read('./report.txt'))

1.times {
  save_report(report)
}

```
I give them json file (report txt) with coverage data that contain hash with around 5000 pairs, from my test environment. This is my results:
**With Redis**
```
@@@existing coverage
  0.984000   0.578000   1.562000 (  1.448660)
@@@array
  0.000000   0.016000   0.016000 (  0.021694)
@@@each
  0.062000   0.000000   0.062000 (  0.076608)
@@@store
  0.532000   0.469000   1.001000 (  0.861668)
```
**With Fake Redis (only ruby overhead)**
```
@@@existing coverage
  0.031000   0.000000   0.031000 (  0.022227)
@@@array
  0.000000   0.000000   0.000000 (  0.000298)
@@@each
  0.078000   0.000000   0.078000 (  0.072801)
@@@store
  0.000000   0.016000   0.016000 (  0.013209)
```

This is too slow. Can you run this benchmark on your system?

**Comment by viktor-silakov** [comment]
I obfuscate my production data for this benchmark and attach to this comment.
[report.txt](https://github.com/danmayer/coverband/files/2286586/report.txt)

**Comment by danmayer** [comment]
OK, this is really interesting I will dig in, I likely won't have time to do that until Thursday. If you could privately email me the list of filenames or give some subset examples that would be helpful. As 5K files seem like a large amount.

 My current theory is that it is tracking ALL gem usage, libraries and system ruby files and that your  `config.ignore = ['lib/scrazy_i18n_patch_thats_hit_all_the_time.rb']` and perhaps the default should be updated to make sure this doesn't happen or that Coverband should at least give a warning. Nearly all my projects for example have:

`config.ignore            = %w[vendor .erb .slim]`

which ignores the libraries / Gem to focus only on application source usage.

The other way to avoid this cost would be to not report the data on every request to your server

changing   

`config.percentage = Rails.env.production? ? 100.0 : 100.0`
to something like
`config.percentage = 5.0`

then while all coverage data is collected it is only reported after 5% of the requests... For this the issue would be when your test suite is done you need a way to force coverband to submit the data before shutting down the test server. I can expose a way to do this via the API.

**Comment by viktor-silakov** [comment]
@danmayer have you already run my performance test? The data (my obfuscate production) i attach to my previous comment.

**Comment by danmayer** [comment]
sorry I haven't gotten a chance to run through your performance test yet, I will try to set it up sometime soon. I am out on vacation until tomorrow so didn't do much code last week ;)

**Comment by danmayer** [comment]
Hmmm OK, so something seems off this might help us figure out the issue my Redis overhead is nowhere near what you are seeing... For your specific use case, I think we can get rid of the Redis overhead entirely, but I would still like to know what is going on.

This is with your fake redis

```
ruby test/benchmarks/redis_speed_test.rb
@@@existing coverage
  0.030000   0.000000   0.030000 (  0.037967)
@@@array
  0.000000   0.000000   0.000000 (  0.000398)
@@@each
  0.080000   0.010000   0.090000 (  0.092125)
@@@store
  0.020000   0.000000   0.020000 (  0.016076)
```

This is with my local Redis server

```
[master][~/projects/coverband] ruby test/benchmarks/redis_speed_test.rb
@@@existing coverage
  0.100000   0.020000   0.120000 (  0.133146)
@@@array
  0.010000   0.000000   0.010000 (  0.018160)
@@@each
  0.070000   0.010000   0.080000 (  0.081094)
@@@store
  0.290000   0.050000   0.340000 (  0.445605)
  ```

My local redis server, could you list the version you are on as we do some optimizations for newer redis when available. Please let me know your info.

```
redis-server --version
Redis server v=3.0.1 sha=00000000:0 malloc=libc bits=64 build=bf58331b4c8133f5
```

I added an endpoint that can trigger Coverband to record coverage. My thought is you can put the `config.percentage  = 1` then at the end of your test suite ensure you make a web request to `/coverage/record_coverage/` and it will push all the data to Redis without slowing down all your test calls.

viktor-silakov:
- Redis server v=3.0.6 sha=00000000:0 malloc=jemalloc-3.6.0 bits=64 build=687a2a319020fa42

- I must listen all 100% of request on my test installation, then i include this data to test running report.

danmayer:
so I am still curious on why your Redis is 10X slower than mine, that seems very odd.

No matter what you set `config.percentage` you are listening to 100% of the request, always are with the `coverage` adapter there is no way not to be capturing all the coverage data. It is just how often you report that to Redis which is the slow operation. The basic thought would be:

* set config percentage to 1%
* run your full test server and test suite
* make sure the very last think on your test suite calls an endpoint like `/record_coverband`
* that endpoint calls `Coverband::Collectors::Base.instance.report_coverage`
* then test suite ends
* you call `rake coverband:coverage` to get the report

this way you would hardly call redis at all during testing, but would capture ALL the code usage.

danmayer:
let me know if that makes sense @viktor-silakov

viktor-silakov:
Thanks! I got your idea. Can i hit `Coverband::Collectors::Base.instance.report_coverage` via  `coverband:coverage` rake task  at the beginning?

kbaum:
One other point I want to mention. In order to avoid the hit on the first coverband enabled request, I added this to application.rb within rails:

```ruby
      config.after_initialize do
        # Warm up the coverband in memory cache so first request does not suffer
        Thread.new { ::Coverband::Collectors::Base.instance.save }
      end
```

danmayer:
In Coverband 3, I am going to try to push a bunch of those kinds of tips into the default Railties, I have some things like that in most of my apps as well, but it has been one of the pieces that makes it a hard tool for folks to configure.

PR #111, opened 2018-08-14
fix type mismatch error

viktor-silakov:
If you have multiple files with one derived `basic`  class and other 'extension' classes without inheritance, when you run baseline task  you can receive error: `TypeError: superclass mismatch for class classname`.
For example two files:
**a1.rb** 
`
class A1 < TracePoint
end`
**a1_ext.rb** 
`class A1
end`
If you load them in order basic -> extension you get TypeError in `safely_import_files` method. To prevent it i change require order and add  Rails.application.eager_load! for rails env.

### DIFF
--- a/lib/coverband/tasks.rb
+++ b/lib/coverband/tasks.rb
@@ -36,14 +36,14 @@ namespace :coverband do
         require baseline_file if File.exist?(baseline_file)
       end
 
-      safely_import_files(Coverband.configuration.additional_files.flatten)
-
       if defined? Rails
+        Rails.application.eager_load!
         safely_import_files(Dir.glob("#{Rails.root}/app/**/*.rb"))
         if File.exist?("#{Rails.root}/lib")
           safely_import_files(Dir.glob("#{Rails.root}/lib/**/*.rb"))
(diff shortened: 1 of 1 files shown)


danmayer:
I see how this could occur, I don't know if I want to add this though as the plan is to remove the baseline functionality entirely. I don't believe it is needed with the `coverage` method of recording as any rails app run with Rails eager loading will already record all the baseline data on app startup. 

Do you hit this error outside of running the `baseline` or just when running the `baseline` command? If you hit it during reporting, collection, or app startup I would be interested to know more. 

I guess have you cleared coverage data and run the coverage collection never running the baseline to see if you get the same results and coverage? If so I think it is another point towards me just removing this functionality entirely.

danmayer:
I changed my mind on this it is definitely a fix that improves the situation now so I will merge it in, but still work towards removing the concept of `baseline` entirely. One issue I think may occur with `baseline` is that folks eager load in production but not in dev, test, or staging environments that often. Which means folks might want baseline while testing or working on Coverband. In which case this would be helpful.

PR #112, opened 2018-08-15
fix simplecov report wrong relevant lines

viktor-silakov:
The relevant lines in the simplecov report is wrong. All unhitten lines coverband makes as irrelevant. I create method-adapter that convert to right scov data.

### DIFF
--- a/lib/coverband/reporters/simple_cov_report.rb
+++ b/lib/coverband/reporters/simple_cov_report.rb
@@ -3,6 +3,18 @@
 module Coverband
   module Reporters
     class SimpleCovReport < Base
+
+      def self.normalize_scov_report(report)
+        report.inject({}) do |mem, (file, cover_hits)|
+          mem.merge(
+              { file => (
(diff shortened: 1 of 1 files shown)


viktor-silakov:
my_fork
coverband
I fix this in my fork, see `fork` branch. This commit is too raw, even my fix in fork is raw, but you can use my code to understand how to create more more thoughtful fix. Test project:
[coverband_test.zip](https://github.com/danmayer/coverband/files/2302003/coverband_test.zip)

danmayer:
I will need to better document this but this is perhaps why I need to kill the trace collector as they function very differently

With `trace` you can turn on and off the collection of coverage and actually sample. With `coverage` collector there is no way to disable coverage once it has started so there is no on an off the only thing you change is the reporting frequency not the sampling. So how often you send data to Redis.

but in this case I get what you are showing which is loading a file and having coverage on it doesn't get the same as if it was loaded directly by simplecov / Coverage standard lib. There was a fix for this elsewhere so I  probably need to work on getting the improvement in that place opposed to fixing at the end on reporting, but on initial file load.

PR #116, opened 2018-09-13
Use correct variable for filestore

danrabinowitz:


danmayer:
ouch thanks missed that in the refactoring, glad you caught it before the next release goes out. I have had that in testing on production. Thanks @danrabinowitz

Issue #115, opened 2018-09-11
Performance issue with Redis Adapter (2.0.1)


CharlesMcMillan:
At one point previously, we used 1.5.4 with success on our code base, and we wanted to reintroduce coverband with the newer version, 2.01...however we are seeing performance issues when coverband is enabled and collecting/sending data to redis... Here is a gist that demonstrates the problem (runnable in a `bundle console` session) https://gist.github.com/CharlesMcMillan/6bc239429761adc70ef738d05e82de6b

The goal of the gist here is to simulate the cost associated with a single page load in our app, which might include something like 10 requests where the 'report' being sent will include 2394 files with 24 lines of code each (hopefully this demystifies some of the number choices in the gist!)

The results indicate that we are losing between 1-1.5 seconds per request (I got that number by dividing 11.558872 / 10), given the above assumptions/guesstimations

It appears that saving a 'report' to redis is pretty costly here. Did something major change between 1.5.4 and 2.0.1 that would explain this?

danmayer:
hmm @CharlesMcMillan see some of my notes on the ticket https://github.com/danmayer/coverband/pull/117#event-1844761550 I am happy to merge that in and test it as I was about to release the next version of Coverband and I could try to get this in first. Take a look and let me and @kbaum know if you think that would help.

I believe it would help the issue you are having but I am also worried that the optimal config settings aren't being used.... could you tell me the config settings you are using (making sure to keep out any host / password details)... Particularly seeing the collector, store, percentage settings.

danmayer:
OK try again with memory store configured and using the 2.0.2 release. 

You can see it in action here https://coverband-demo.herokuapp.com/
and the config enabling memory caching here https://github.com/danmayer/coverband_demo/blob/master/config/coverband.rb

PR #117, opened 2018-09-13
Memory cache store revamp

kbaum:
This PR updates the MemoryCacheStore to work with the latest +2.x changes to coverband. The memory cache store filters out files with unchanging coverage from being sent to the underlying store (redis).  

This update resolves https://github.com/danmayer/coverband/issues/115.

To demonstrate using the [code](https://gist.github.com/CharlesMcMillan/6bc239429761adc70ef738d05e82de6b) from the issue, we can just wrap the redis_store with the memory_cache_store:

```ruby
report = 2934.times.each_with_object({}) do |file_number, hash|
  line_numbers = 24.times.each_with_object({}){ |line, hash| hash[(line+1).to_s] = 0 }
  hash["file#{file_number+1}.rb"] = line_numbers
end
store = Coverband::Adapters::MemoryCacheStore.new(Coverband::Adapters::RedisStore.new(Redis.new))
5.times { store.save_report(report) }
bm = Benchmark.measure do
  10.times do
    store.save_report(report)
  end
end
```
Running the above, yields a benchmark of:

```
0.016358   0.000016   0.016374 (  0.016382)
```

Which works out to 99.86% improvement from the benchmark referenced in https://github.com/danmayer/coverband/issues/115.

### DIFF
--- a/lib/coverband/adapters/memory_cache_store.rb
+++ b/lib/coverband/adapters/memory_cache_store.rb
@@ -9,16 +9,18 @@ module Coverband
     class MemoryCacheStore < Base
       attr_accessor :store
 
+      @@files_cache = {}
+
       def initialize(store)
         @store = store
       end
 
-      def self.reset!
-        files_cache.clear
+      def self.clear!
+        @@files_cache.clear
       end
 
       def clear!
-        self.class.reset!
+        self.class.clear!
       end
 
       def save_report(files)
@@ -26,28 +28,26 @@ module Coverband
         store.save_report(filtered_files) if filtered_files.any?
       end
 
-      # rubocop:disable Lint/IneffectiveAccessModifier
       private
 
-      def self.files_cache
-        @files_cache ||= {}
-      end
-
       def files_cache
-        self.class.files_cache
+        @@files_cache
       end
 
       def filter(files)
-        files.each_with_object({}) do |(file, lines), filtered_file_hash|
-          # first time we see a file, we pre-init the in memory cache to whatever is in store(redis)
-          line_cache = files_cache[file] ||= Set.new(store.covered_lines_for_file(file))
-          lines.reject! do |line|
-            line_cache.include?(line) ? true : (line_cache << line && false)
+        files.each_with_object({}) do |(file, covered_lines), filtered_file_hash|
+          if covered_lines != cached_file(file)
+            files_cache[file] = covered_lines
+            filtered_file_hash[file] = covered_lines
           end
-          filtered_file_hash[file] = lines if lines.any?
         end
       end
-      # rubocop:enable Lint/IneffectiveAccessModifier
+
+      def cached_file(file)
+        files_cache[file]  ||= store.covered_lines_for_file(file).each_with_object({}) do |(line_number, value), hash|
+          hash[line_number.to_i] = value.to_i
+        end
+      end
     end
   end
 end
--- a/test/unit/adapters_memory_cache_store_test.rb
+++ b/test/unit/adapters_memory_cache_store_test.rb
@@ -5,52 +5,62 @@ require File.expand_path('../test_helper', File.dirname(__FILE__))
 module Coverband
   class MemoryCacheStoreTest < Test::Unit::TestCase
     def setup
-      Adapters::MemoryCacheStore.reset!
-      @store = mock('store')
+      Adapters::MemoryCacheStore.clear!
+      @redis = Redis.new
+      @store = Coverband::Adapters::RedisStore.new(@redis)
+      @store.clear!
       @memory_store = Adapters::MemoryCacheStore.new(@store)
     end
 
-    def data
-      {
-        'file1' => { 3 => 1, 5 => 2 },
-        'file2' => { 1 => 1, 2 => 1 }
-      }
-    end
-
     test 'it passes data into store' do
+      data = {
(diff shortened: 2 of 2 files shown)


danmayer:
hmm interesting I was considering removing this cache as I feel like in ideal usage, one shouldn't be running into reporting issues... Perhaps, I should look at allowing this and some other ways to reduce the impact for folks switching from the old tracepoint coverage to the new one... as it has several differences...

* previously one needed to have a reasonble high coverband percentage to get valuable data... moving to `coverage` one should have a very low percentage just something to ensure one is reporting at checkpoints before deployments or webserver restarts. (0.01% is what I run in prod now) 
           * perhaps I should change the dev mode default, as it recommends 100% which is useful while setting up and debugging but after that un-wise.
* coverband used to only record and send up app files but coverage by default caputres far more library code... perhaps the default should be to ignore anything but app code unless folks set a setting in coverband... `config.track_external` as this did balloon the number of files being tracked and sent to Redis significantly for me.

basically this makes me think opposed to using this memory cache one should just be adjusting the reporting period to something that makes sense. Although this cache is a simple way to avoid calls on many frequerntly unused files or config files just loaded during initialization and in theory would never really slow things down.

danmayer:
Thanks @kbaum added a comment on that issue and will try to do some testing to get this in the next release which will likely go ou this weekend.

kbaum:
Full disclosure: @CharlesMcMillan work at the same company.

That is interesting about not needing a high coverage percentage. I hadn't thought about the fact that since we are using ruby Coverage, we are always tracking coverage for 100% of requests even if we are not reporting to the store for that request.

RE: Low percentage of reporting. The problem we are noticing is that because we have a very large code base, when we do report, coverband makes thousands of calls to redis. It is very slow when we are running redis on the same machine, but that same benchmark is way slower when running on heroku using heroku redis. Running the same benchmark referenced in issue 115 yields the following results:

```
31.536000   8.288000  39.824000 (126.144325)
```

So when coverband is enabled for a request, we are adding another 12.6 seconds on average.

danmayer:
I had guessed y'all were at the same place but didn't want to jump to conclusions, thanks for the transparency... Also, cool to see someone using for so long ;)

Wow! ok, thanks I was also not understanding that use case (in issue 115) and we had been back and forth...

I think this makes two things clear...

* I need an entirely different config option for `coverage` vs tracepoint so folk understands the implications... one really was how many samples you get and one was really how often we report that. So splitting to two clear config options is best.
* That your patch will still be helpful as the project in general needs to reduce Redis calls
* by default I should make sure that the gem ONLY sends app files and ignores lib / gem / system ruby stuff unless user opt into it (right now you need to use Coverband ignore to opt OUT of it)

Does that all make sense to you @kbaum

kbaum:
That all makes sense. Is there a reason to keep tracepoint around if we can reduce the calls to redis? From the benchmarks I’ve run, i may be jumping to conclusions but it seems like ruby coverage has very little overhead. I also wonder if it makes sense to think about running a background thread which periodically stores coverage information in redis. Then users would really not incur much of a performance hit when reporting on numerous files perhaps including gems.

danmayer:
yeah see recent blog posts and such but the overhead of running Coverage all the time now is extremely low... back when I build the tracepoint that wasn't the case and `peek_results` didn't exist, which really enabled using it.

yeah @kbaum the project should probably drop Tracepoint I have that on the change list for the future 3.0 release

https://github.com/danmayer/coverband/blob/master/changes.md#coverband-30

I didn't do it right away thinking:

* this would make an easier upgrade path for folks, from the various feedback I don't think that has been true, as there were enough differences it was confusing
* I still see some use cases for Tracepoint that I am exploring...
     * by not having the sampling / reporting together, the project couldn't do things like tag coverage just for the currently executing route
     * allow only to collect coverage based on route
     * coverage over some other variable like a set of alpha users
     * all of these tracepoints can do, but STDLib `Coverage` can't... which could be an interesting tradeoff for some use cases... I can't decide if any are compelling enough to keep it in.

Anyways, any feedback on those ideas appreciated.

kbaum:
I might be biassed, but I don't think the use cases for tracepoint are compelling enough to keep around when I compare to the benefits in simplicity in removing it. For one, I think we could really simplify the experience for the users in setting this up. Less options means less to think about. Just drop in the gem, point to your redis instance, and it just works. If we completely remove the performance overhead, then theres no need to configure anything else. At that point coverband feels like a no brainer for everyone.

I've also gotten pretty familiar with the code again over the last week and removing tracepoint and sticking with coverage could really simplify it. The logic could be just enabling coverage and kicking of a thread that reports to redis every 30 seconds. No need for any middleware or different implementations of collectors.

danmayer:
OK let me get this merged in, and get 2.0.2 out as it resolves some issues... Then I will start work on the 3.0 branch and dump tracepoint. I agree that making Coverband easier to use is probably the biggest way to increase adoption, it is still to easy to shoot yourself in the foot with the tool and cause performance issues if you don't tune it well.

I think given all the server options and issues with pre-post fork, I might keep around the middleware as optional, but as I add in the threading, I guess we will see how well it works for all cases. I have been considering a background method for awhile as the performance issue shifted from collecting code usage to reporting it.

I will ping you to take a look at the work in progress 3.0 branch when I get it kicked off.

kbaum:
Great. Thanks @danmayer!

Issue #124, opened 2018-09-28
Background Coverage Data Reporting


Kallin:
We're trying out coverband on production, and finding that the redis calls are taking a majority of the time of some of our web requests when looking at new relic. 

Digging into the redis_store, it seems that, for every file, we first get all of the existing values using `existing = redis.hgetall(key)`, then after incrementing the line counts, we use `redis.mapped_hmset(key, values)` to update redis.

In our application this results in hundreds of gets, and hundreds of sets to redis in the middleware. I have a couple ideas and I'm curious what the devs here think:

1. Fork this call, https://github.com/danmayer/coverband/blob/master/lib/coverband/middleware.rb#L14 . The web request doesn't need to wait for anything from this report generation, so why not let it run in another process?

2. Pipeline the calls instead of doing things one file at a time. In our case, that would mean 2 redis commands instead of ~300 on some requests. 

ie, here: https://github.com/danmayer/coverband/blob/master/lib/coverband/adapters/redis_store.rb#L26 ,  we iterate through every file in the report, and for each we get the existing values then increment and update here: https://github.com/danmayer/coverband/blob/master/lib/coverband/adapters/redis_store.rb#L53 . It looks like redis supports pipelining, https://redis.io/topics/pipelining, which would allow all these requests to be grouped together.  I've tried this and it seems to work:

```ruby
def save_report(report)
  store_array(base_key, report.keys)

  keys = report.keys.map {|file| "#{base_key}.#{file}"}

  existing_records = redis.pipelined do
    keys.each do |key|
      redis.hgetall(key)
    end
  end

  combined_report = {}
  keys.each_with_index do | key, i|
    combined_report[key] = {
      new: report.values[i],
      existing: existing_records[i]
    }
  end

  redis.pipelined do
    combined_report.each do |file, values|
      existing = values[:existing]
      new = values[:new]
      unless values.empty?
        # in redis all keys are strings
        new_string_values = Hash[new.map { |k, val| [k.to_s, val] }]
        new_string_values.merge!(existing) { |_k, old_v, new_v| old_v.to_i + new_v.to_i }
        redis.mapped_hmset(file, new_string_values)
        redis.expire(file, @ttl) if @ttl
      end
    end
  end

  # report.each do |file, lines|
  #   store_map("#{base_key}.#{file}", lines)
  # end
end
```

[...]

PR #125, opened 2018-10-01
allowing reporting to happen in forked process

Kallin:
simple forking solution to support performance concerns from https://github.com/danmayer/coverband/issues/124

### DIFF
--- a/lib/coverband/collectors/base.rb
+++ b/lib/coverband/collectors/base.rb
@@ -54,6 +54,7 @@ module Coverband
         @store = Coverband::Adapters::MemoryCacheStore.new(@store) if Coverband.configuration.memory_caching
         @verbose  = Coverband.configuration.verbose
         @logger   = Coverband.configuration.logger
+        @fork_reporting   = Coverband.configuration.fork_reporting
         @current_thread = Thread.current
         Thread.current[:coverband_instance] = nil
         self
@@ -76,8 +77,20 @@ module Coverband
         raise 'abstract'
       end
 
-      def report_coverage
+      def report_coverage_potentially_forked
         raise 'abstract'
+      end
+
+      def report_coverage
+        if @fork_reporting
+          pid = Process.fork do
+            puts 'reporting in fork'
+            report_coverage_potentially_forked
+          end
+          Process.detach pid
+        else
+          report_coverage_potentially_forked
+        end
       end
 
       protected
--- a/lib/coverband/collectors/coverage.rb
+++ b/lib/coverband/collectors/coverage.rb
@@ -11,7 +11,7 @@ module Coverband
         # noop
       end
 
-      def report_coverage
+      def report_coverage_potentially_forked
         unless @enabled
           @logger.info 'coverage disabled' if @verbose
           return
--- a/lib/coverband/collectors/trace.rb
+++ b/lib/coverband/collectors/trace.rb
@@ -39,7 +39,7 @@ module Coverband
         unset_tracer
       end
 
-      def report_coverage
+      def report_coverage_potentially_forked
         unless @enabled
           @logger.info 'coverage disabled' if @verbose
           return
--- a/lib/coverband/configuration.rb
+++ b/lib/coverband/configuration.rb
@@ -7,7 +7,7 @@ module Coverband
                   :reporter, :startup_delay, :memory_caching,
                   :include_gems, :collector, :disable_on_failure_for,
                   :redis_namespace, :redis_ttl,
-                  :safe_reload_files
+                  :safe_reload_files, :fork_reporting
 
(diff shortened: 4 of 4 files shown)


Kallin:
please let me know if you'd like to see this done differently or if there's anything else I can do to help get this PR accepted.

kbaum:
This is a very interesting idea. My only thought is that we might get the performance we need by just reporting to redis in a background thread instead:

https://github.com/danmayer/coverband/issues/124#issuecomment-425582122

Then we avoid the added branching logic in the code.

lib/coverband/collectors/base.rb
```diff
@@ -76,10 +77,22 @@ def stop_coverage
         raise 'abstract'
       end
 
-      def report_coverage
+      def report_coverage_potentially_forked
```
kbaum: To avoid the extra "abstract" method, can we pull the forking logic up to the base class?

kbaum: Ok. I think i misread this initially. Apologize for that.

kbaum:
RE: Sampling. With the ruby coverage approach to coverage as opposed to trace, coverage is being tracked all of the time.. whether we report for that request or job execution. Sampling percentage then only controls how often we report to redis. Since the coverage data is global available to the process through [Coverage#peek_result](https://ruby-doc.org/stdlib-2.5.0/libdoc/coverage/rdoc/Coverage.html#method-c-peek_result), there is no need to report to redis during the request.

RE: Threads and ruby. It is true that ruby cannot perform concurrent processing across multiple CPU's. That being said, when one thread is waiting on IO, another thread of execution can run.  This is an old post but still holds up:

https://yehudakatz.com/2010/08/14/threads-in-ruby-enough-already/

In the case of reporting to redis, all we are doing is IO so I would think that this would be an excellent case to leverage threads.

danmayer:
seems lots of folks are interested in the same basic idea, I have started work on coverband 3 to pull things out and simplify so I don't think we want many PRs on the older code base before that... I think forking is interesting and I plan to keep a percentage based reporting and not move entirely to the background thread as I agree there are a number of use cases where it will still be an easier solution... The other thing is really how we are reporting now the thousands of redis calls isn't a good use a single store hash and summing across them only at report time is probably more efficient.

danmayer:
I pulled in the pipeline change, but would like to hold off on these kinds of concepts until `coverband 3` as I think we will be able to significantly improve reporting once simplified

Kallin:
Thanks for taking a look, and I look forward to seeing what shows up in Coverband 3.

Re: a couple of your comments:

Forking & Memory usage: I suppose this may depend on the underlying OS, but my understanding is that Linux and modern OSes implement copy-on-write when forking in order to keep forked processes lightweight, https://unix.stackexchange.com/questions/155017/does-fork-immediately-copy-the-entire-process-heap-in-linux . I suppose if we wanted to verify this we could create a bunch of ruby forks that just sit around sleeping for a while and keep forking and see how memory usage increases with each fork.

It would be interesting to compare performance of forking w/ background thread in a 'standard' scenario (I'm thinking rails web request). Forking should add close to no overhead to the time of the request, as the request just forks and never waits. Threading will add some overhead I imagine due to context-switching, but it's hard to know how much. If any of the 'reporting' thread takes a chunk of time other than in low-level IO, the web request would be waiting on it. I wonder what % of the time spent in the reporting is actually low-level IO, and how much is spent in the redis library marshalling/unmarshalling request/response etc., that would not qualify as IO and would block other threads.

It may be a low overhead with the threading, but in our case we have a very strict time limit on certain requests that we want to keep things under, so even adding 100ms or something to a request may be too much. 

Anyways, I'd encourage a deeper look at forking, I'm still bullish on it :)

kbaum:
You make some very compelling points re: forking. One thought is that with the new coverage collector, there is no need to report to redis during a request since coverage data is global for the process. We could even have a background thread that loops, sleeps, forks, and reports.

danmayer:
Yeah I will have to do some tests but my thinking would be for a multi-threaded server you would have the parent process start the thread and do reporting out of request cycle... 

I think the most common target is Puma, which is interesting because of cluster mode...

`Puma also offers "clustered mode". Clustered mode forks workers from a master process. Each child process still has its own thread pool.`

If we can support background reporting in clustered mode across processes and threads I think that should cover all the various use cases... Then we just need to find the various hooks for alternative servers or offer a middleware / forking option for folks on non-standard servers.

kbaum:
I noticed that when unicorn and puma fork, ruby coverage is reset for some reason.
https://github.com/danmayer/coverband/issues/134
Would that apply here?

kbaum:
@Kallin - going to close this for now as this has some conflicts and I am seeing that forking resets coverage in the forked process. We can always reopen once conflicts are resolved if i am wrong on that.

PR #127, opened 2018-10-08
Filter out files with no coverage

kbaum:
With rails eager loading on, I notice that 95-99% of the files being reported to redis have no coverage data but we are still sending requests to redis for these files. I have a hunch that this will solve many people's performance issues after the first request.

### DIFF
--- a/lib/coverband/collectors/coverage.rb
+++ b/lib/coverband/collectors/coverage.rb
@@ -36,7 +36,7 @@ module Coverband
         end
 
         if @store
-          @store.save_report(@file_line_usage)
+          @store.save_report(files_with_line_usage)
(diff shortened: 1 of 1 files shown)


danmayer:
excellent this is great thanks @kbaum I think you are right that this will help many real world use cases.

danmayer:
I was hoping one of my benchmarks would be able to prove a win here, but no such luck... I believe my AB_Benchmark method against a full rails app would show a big win here, but I haven't set that up to be easily repeatable yet. I am thinking a benchmark script in the coverband demo app would be a good way to prove the perf gains seen in standard rails apps. I will try to add something like that in the next few days and report back. Thanks again @kbaum this should be a big win.

Issue #118, opened 2018-09-17
Include checksum as part of key for each file in store


kbaum:
This is more of an idea than an issue. If we include the checksum as the key for each file within the redis store, we would remove the need for the user clearing the store. For example:

```ruby
Digest::MD5.file('app/models/user.rb').hexdigest
```

This would also remove confusion around a files coverage not being accurate when the file has been changed but the store has not been cleared.

In order to avoid stale files hanging out forever within redis, we could also just supply a high default ttl and the stale files should automatically be cleaned up by redis.

danmayer:
interesting idea, we do have a ttl option but it defaults to forever... so something like 1 year as a high default would work...

I like the idea, I had been considering namespacing by deployed git hash for a while, but this in some ways is better as it doesn't impact files that weren't changed between deployments.

danmayer:
updated title from "Include checksum as part of key for each file in store"

danmayer:
I rephrased this as I think there are a couple different things that could be helpful.

One thing was that `clearing` Coverband coverage can be confusing as people don't know or trust how long the data has been there... The most often reason for clearing is because of the coverage drift where line numbers don't match the captured data anymore.

While the checksum solution solves the drift issue, it does cause a slightly different trust issue. Basically, that a frequently changed file might no have a very long history showing usage... So it could look like it isn't used if it was just part of a not often run cron job or something like that.

I guess in a perfect world we would really have line number level MD5 and only throw out coverage when the actual line has moved / changed. This is not realistic.

Ideas to help with understanding the time range that the coverage data is valid for, and to avoid the drift.

* store some metadata like when the file first started to record coverage data...
* tie this with the MD5 idea

update the UI so for any file you would have something like

```
app/controllers/users_controller.rb version XYZ first seen on 09-15-2018 08:05am, last coverage updated on 10-03-2018 08:05am
```

In theory, one could actually users see all versions of a file and navigate back and forth... seeing coverage change over file history... over being able to generate a coverage report based on time range... Show me coverage data between date X and Y.

I had been considering similar ideas based on git hashes at time of deploy... so you could query coverage based on related git versions or release time... but I do think the MD5 version has advantages as it wouldn't be tied to any code repository or deployment methods.

The benchmarks you shared @kbaum make it seem reasonably fast... I think when first introduced it still probably makes sense to do as an optional feature.

[...]

PR #128, opened 2018-10-08
Include checksum as part of key for each file in store

dnasseri:
Fixes: https://github.com/danmayer/coverband/issues/118

This PR adds a checksum key to each file in redis. The checksum is computed with with the contents of the file. Before persisting to redis, we determine if the checksum has changed. If so we overwrite the values for the given file in redis, if not we merge them.

Was speaking with @kbaum and we are thinking it makes sense to add a cache for the checksum so we don't have to compute the hash every time.

### DIFF
--- a/lib/coverband/adapters/redis_store.rb
+++ b/lib/coverband/adapters/redis_store.rb
@@ -5,10 +5,19 @@ module Coverband
     class RedisStore < Base
       BASE_KEY = 'coverband2'
 
+      attr_accessor :checksum_generator
+
+      class ChecksumGenerator
+        def generate(file)
+          Digest::MD5.file(file).hexdigest
+        end
+      end
+
       def initialize(redis, opts = {})
         @redis = redis
         @ttl             = opts[:ttl]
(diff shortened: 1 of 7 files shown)


danmayer:
OK this feature I think we should likely hold until version 3.0. Cool to work on and try out the proof of concept but likely will have some conflicts / rewriting when the 3.0 branch gets completed. That make sense @dnasseri / @kbaum

kbaum:
@danmayer - I trust your judgement here but I do think this would be a big leap forward in making coverband a low maintenance library for users while improving it's trustworthiness. Currently users have to figure out a strategy for how often to clear their data for all of their source files. In a decent size team, other developers start to doubt the coverage data because they are unsure how long coverband has been gathering data. With this change, coverband users can just let coverband track forever. I was thinking we should follow up this change with added information on the report about when coverband began tracking data for this file. For example:

Coverage data for user.rb since 9/25/2018

danmayer:
ok closing this out as the work and ideas are being pulled into https://github.com/danmayer/coverband/pull/137

PR #129, opened 2018-10-15
Feature/coverband 3

danmayer:
OK, I expect to do a bit more cleanup, but relative paths work, tests all pass, and benchmarks are smoking fast. I have Coverband demo running off the alpha branch locally without issue.

Take a look when you have some time... I will push further cleanup on this branch, I will likely collapse the commits down to something more readable for the final PR, but figured it was in good enough shape for some feedback/discussion.


changes.md
```diff
@@ -4,22 +4,39 @@
...
-   * expects to drop the concept of baseline recordings
-   * improve support for eager-loading
-   * add built-in support for easy loading via Railties
-   * expects to add safe list support to force reload files one wants coverage on that may happen outside of the standard load order
-   * built in support for activejob, sidekiq, and other common frameworks
-* code route tracing (entry point to all code executed for example /some_path -> code coverage of that path)
+- expects to drop Tracepoint collection engine
+- expects to drop anything below Ruby 2.3
+- Release will be aimed as significantly simplifying ease of use
+  - near zero config setup for Rails apps
```
kbaum: 👏

lib/coverband/adapters/file_store.rb
```diff
@@ -46,6 +42,10 @@ def covered_lines_for_file(file)
 
       private
 
+      def save_coverage(report)
+        File.open(path, 'w') { |f| f.write(report.to_json) }
```
kbaum: By default use file system to persist report?

lib/coverband/adapters/s3_report_writer.rb
```diff
@@ -0,0 +1,59 @@
...
+      def persist!
+        object.put(body: coverage_content)
+      end
+
+      private
+
+      def coverage_content
+        version = Gem::Specification.find_by_name('simplecov-html').version.version
+        File.read("#{SimpleCov.coverage_dir}/index.html").gsub("./assets/#{version}/", '')
+      rescue StandardError
```
kbaum: Why do we need a rescue here?

lib/coverband/adapters/s3_report_writer.rb
```diff
@@ -0,0 +1,59 @@
...
+
+module Coverband
+  module Adapters
+    ###
+    # TODO: this is currently a html report writer
+    # this should support writing coverage the current method should be fine
+    # to write every report to S3 and sum them later or use the 2 pass
+    # method we do for redis if in a background thread
+    ###
+    class S3ReportWriter
```
kbaum: Is this class an adaptor like RedisStore?

lib/coverband/adapters/file_store.rb
danmayer: so this is only for the file store which mostly has been used for testing and development it would work fine for single server deployments but falls over for anything with more than one server so Redis is the primary adapter... but with the new pattern I think it would be simple to support memcache, ActiveRecord, and S3 as a store... I think for S3 just logging all the files and doing the combinational math to remove duplicate counts only when reporting

lib/coverband/adapters/s3_report_writer.rb
danmayer: yeah I feel like I might move the S3 html reporter to utils and make a pure S3 adapter

lib/coverband/adapters/s3_report_writer.rb
danmayer: cool call let me fix

lib/coverband/adapters/file_store.rb
kbaum: Oh gotcha. I think i misread this before. Makes sense.

lib/coverband/adapters/s3_report_writer.rb
kbaum: Wouldn't this need a save_coverage method to be an adaptor?

Kallin:
@danmayer , great, if you decide to publish an alpha gem, I can try deploying it in some test environments at work to see if there are any issues here.

lib/coverband/adapters/s3_report_writer.rb
danmayer: yeah let me move this out of adapters until I actually implement those pieces and instead have what this does live in Utils.

danmayer:
cool thanks @Kallin I want to do a bit more cleanup and I am headed out of town, but I should be able to get an alpha release live early next week... Monday or Tuesday...

lib/coverband/collectors/base.rb
```diff
@@ -10,7 +10,7 @@ def self.instance
         if Coverband.configuration.collector == 'coverage'
           Thread.current[:coverband_instance] ||= Coverband::Collectors::Coverage.new
         else
-          raise 'select valid collector [trace, coverage]'
+          raise 'select valid collector [coverage]'
```
kbaum: Should we just not have this be configurable if there's only one choice?

lib/coverband/adapters/s3_report_writer.rb
danmayer: I was going to fix this but unfortunately, I can't remember the case and condition that caused this fallback. I think I will have to leave it until I can figure out a way to recreate the issue.

lib/coverband/adapters/s3_report_writer.rb
danmayer: moved

lib/coverband/collectors/base.rb
danmayer: cleaned up collector base is gone

danmayer:
OK I still need to update a bunch of the docs in the README and such, but I have the first version of coverband_3 alpha released gem up and running the coverband demo site. https://coverband-demo.herokuapp.com/ you can feel free to give it a shot @Kallin if you like see this repo for how to configure and setup in your application.rb https://github.com/danmayer/coverband_demo

Given all the perf improvements and reduced complexity, I am thinking I might make the first release of Coverband 3 before I get all the Railsties stuff in...I think that could be added as a feature release like 3.1 or if there are configuration changes make it Coverband 4. Thoughts on that @kbaum 

I will do some additional test and put it into some other rails sites tomorrow

kbaum:
I think it  makes sense to wait on the Railties stuff since it will take more testing.

changes.md
```diff
@@ -4,22 +4,39 @@
...
+  - add built-in support for easy loading via Railties
+  - built in support for activejob, sidekiq, and other common frameworks
+  - reduced configuration options
+- options on reporting
+  - background reporting
+  - or middleware reporting
+- improved web reporting
+  - no longer relying directly on HTML in S3 but dynamically generated from any adapter
+- additional adapters including Memcache, S3, and ActiveRecord
```
kbaum: I'm feeling like we should be opinionated and just stick with redis. More options just means more confusion.
* Memcached doesn't persist.
* S3 and ActiveRecord are too slow.

lib/coverband/reporters/base.rb
```diff
@@ -2,25 +2,26 @@
...
+          msg = "report:\n #{scov_style_report.inspect}"
+          Coverband.configuration.logger.debug msg
         end
         scov_style_report
       end
 
-      def self.get_roots
+      # protected
+      # below are all not public API
+      # but as they are class methods protected doesn't really do anything
```
kbaum: Can we use private_class_method declaration?

https://apidock.com/ruby/Module/private_class_method

lib/coverband/reporters/base.rb
```diff
@@ -30,19 +31,26 @@ def self.current_root
...
       def self.fix_file_names(report_hash, roots)
-        fixed_report = {} # normalize names across servers
-        report_hash.each_pair do |key, values|
+        if Coverband.configuration.verbose
+          Coverband.configuration.logger.info "fixing root: #{roots.join(', ')}"
+        end
+
+        fixed_report = {}
+        # normalize names across servers
+        report_hash.each_pair do |key, vals|
```
kbaum: Minor.. could get rid of the fixed_report local variable and just use each with object here?

```ruby
report_hash.each_with_object({}) do |(key, vals), fixed_report|
```

lib/coverband/reporters/base.rb
```diff
@@ -82,64 +76,25 @@ def self.filename_from_key(key, roots)
...
         scov_style_report = {}
-
-        ###
-        # why do we need to merge covered files data?
-        # basically because paths on machines or deployed hosts could be different, so
-        # two different keys could point to the same filename or `line_key`
-        # this logic should be pushed to base report
-        ###
-        store.covered_files.each do |key|
+        store.coverage.each_pair do |key, line_data|
```
kbaum: Same idea with `each_with_object`.. just a style preference though... No biggie.

lib/coverband/utils/s3_report_writer.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+module Coverband
```
kbaum: It looks like we are still writing scov report to s3 and reporting from their? Would the idea be to have a rake task run periodically that takes data from redis, generates a report, and writes report to s3 as we do for coverband2?

lib/coverband/reporters/web.rb
```diff
@@ -9,71 +9,74 @@ class Web
...
       def call(env)
         @request = Rack::Request.new(env)
 
         if request.post?
           case request.path_info
-          when /\/collect_update_and_view/
+          when %r{\/collect_update_and_view}
             collect_update_and_view
-          when /\/clear/
+          when %r{\/clear}
```
kbaum: Not s show stopper but it doesn't feel right that everyone on that team that can view the coverage report can also clear all the data. Right now it's difficult to know how much to believe in the coverage data because we don't know how long it's been collecting the information.

Soon I think we should merge in the change that uses a file's checksum to automatically reset coverage data on a file by file basis. Then i think we can remove the idea of clearing the data periodically from redis.

changes.md
danmayer: I updated the changes to move some stuff to Coverband 4.0 like railties I changed working on this as 3.0.0 will likely launch with just the Redis adapter, but I still think others would be good to add. As projects not Redis is one of the most common complaints I have heard about using coverband.

* Memcache doesn't persist but could be OK depending on use case, especially if we can show time range that coverage covers (addressing another issue you mention later)
* S3 might be to slow, but if we get the background thread as an option should be fine
* I will need to get some benchmarks but with the new data format, I think ActiveRecord could be nearly fast as Redis, at least a totally viable runtime option

That being said I think we can work for a zero-config Redis option out of the box.

lib/coverband/reporters/base.rb
danmayer: interesting I will look into this

lib/coverband/reporters/base.rb
danmayer: done

lib/coverband/utils/s3_report_writer.rb
danmayer: Yeah so I want to make the web version able to just read from the store and render the HTML, but given some of the issues on 2.x I am not sure I will get that into the initial 3.0.0 release but it could likely be added during the 3.0.0 lifecycle without breaking changes.

I don't really use the `rake coverband:coverage` task anymore and mostly use the web to click generate and load report. I think we can leave both as an option.

lib/coverband/reporters/web.rb
danmayer: yeah good call. I understand the confusion around how long / valid is the data.

1. I think we can add an option to disable this feature, then the rake task would have to be used to clear. I think that can land on the 3.0.x line
2. Agreed on getting to a checksum or git release tag type solution. My biggest concern is around the performance of calculating all the hashes on the fly. Might be able to get in 3.0.x or might make that one of the major features to drive 4.0

kbaum: RE: Performance of checksum. I did some really basic benchmarking calculating user.rb file checksum 1000X and it's very fast:

```
irb(main):006:0> puts Benchmark.measure { 1000.times { Digest::MD5.file('./app/models/user.rb').hexdigest } }
  0.020000   0.010000   0.030000 (  0.027000)
```
That's 0.000027  seconds per md5. In addition, you probably only need to calculate the md5 one time in production as source file md5's do not change after the process starts. That makes it even faster:

```
irb(main):007:0> checksum_cache = {}
=> {}
irb(main):008:0> puts Benchmark.measure { 1000.times { checksum_cache['./app/models/user.rb'] ||= Digest::MD5.file('./app/models/user.rb').hexdigest } }
  0.000000   0.000000   0.000000 (  0.001463)
=> nil
irb(main):009:0>
```
Which is 0.000001463 seconds per md5.

At those numbers, i wonder if it makes sense to fold this feature into 3.0.x? It really takes a major complication away.

lib/coverband/utils/s3_report_writer.rb
kbaum: Interesting.. i never use the web to generate and have the rake report generation running hourly in a heroku scheduler.

lib/coverband/reporters/web.rb
danmayer: OK I will move that back to the 3.0.x so long as we don't have any API breaking changes. Thinking still after 3.0.0.

lib/coverband/utils/s3_report_writer.rb
danmayer: ha interesting to see different use cases, when I started to add the various dynamic web options I have basically stopped using the rake tasks. I think it is worth maintaining both interfaces though. Rake tasks also integrate well into folks CI workflows.

lib/coverband/reporters/base.rb
danmayer: fixed

danmayer:
OK, @Kallin or @kbaum could anyone else run the second alpha release gem, I have it on two apps if someone else could confirm I think it is ready for release. I have updated all the readme docs to match the 3.0 methods

`gem 'coverband', '>= 3.0.0.alpha2', require: false`

you will also need to update the config, application.rb, how the rake tasks are included... any likely any custom code around background jobs or crons... 

Coverband Demo currently shows both old 2.x and 3.x setups, which should make it easy to switch over.

config: https://github.com/danmayer/coverband_demo/blob/master/config/coverband.rb
application.rb: https://github.com/danmayer/coverband_demo/blob/master/config/application.rb#L37
rake: https://github.com/danmayer/coverband_demo/blob/master/Rakefile#L15

PR #137, opened 2018-11-11
use md5 to avoid coverage drift on changes

danmayer:
OK @dnasseri and @kbaum this takes the ideas from https://github.com/danmayer/coverband/pull/128 but pulls them into Coverband 3 and pushes as much of the logic to the adapter base class, as we don't want to have to do much specific to any given adapter.

I believe this should resolve the drift issue: https://github.com/danmayer/coverband/issues/118

* I need to do a bit more work and review of the benchmarks as I think this does slow things down.
* The report doesn't yet show the additional time information, about the timerange valid for a given file, that will take a bit more work and another PR, but I think we could likely get this in without that. 
     * I am thinking to do that I will need to drop direct dependencies on Simplecov / Simplecov-html

### DIFF
--- a/lib/coverband/adapters/base.rb
+++ b/lib/coverband/adapters/base.rb
@@ -4,36 +4,78 @@ module Coverband
   module Adapters
     class Base
       def initialize
-        raise 'abstract'
+        @file_hash_cache = {}
       end
 
       def clear!
         raise 'abstract'
       end
 
-      def save_report(_report)
-        raise 'abstract'
+      # Note: This could lead to slight race on redis
+      # where multiple processes pull the old coverage and add to it then push
+      # the Coverband 2 had the same issue,
+      # and the tradeoff has always been acceptable
+      def save_report(report)
+        data = report.dup
+        merge_reports(data, get_full_report)
+        save_coverage(data)
       end
 
       def coverage
-        raise 'abstract'
+        simple_report(get_full_report)
       end
 
       def covered_files
-        raise 'abstract'
+        coverage.keys || []
       end
 
-      def covered_lines_for_file(_file)
-        raise 'abstract'
+      def covered_lines_for_file(file)
+        coverage[file] || []
       end
 
       protected
 
+      def save_coverage
+        raise 'abstract'
+      end
+
+      def get_report
+        raise 'abstract'
+      end
+
+      def get_full_report
+        get_report
+      end
+
+      def file_hash(file)
+        if @file_hash_cache[file]
+          @file_hash_cache[file]
+        else
+          @file_hash_cache[file] = Digest::MD5.file(file).hexdigest
+        end
+      end
+
+      def expand_report(report)
+        report.each_pair do |key, line_data|
+          extended_data = {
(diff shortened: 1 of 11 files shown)


lib/coverband/adapters/base.rb
```diff
@@ -4,36 +4,78 @@ module Coverband
...
         raise 'abstract'
       end
 
-      protected
+      def get_full_report
+        get_report
+      end
+
+      def file_hash(file)
+        if @file_hash_cache[file]
```
kbaum: Little more concise?
```ruby
def file_hash(file)
  @file_hash_cache[file] ||= Digest::MD5.file(file).hexdigest
end
```

lib/coverband/adapters/base.rb
```diff
@@ -4,36 +4,78 @@ module Coverband
...
         raise 'abstract'
       end
 
-      def covered_lines_for_file(_file)
+      def get_report
         raise 'abstract'
       end
 
-      protected
+      def get_full_report
```
kbaum: Do we still need but get_full_report and get_report?

lib/coverband/adapters/base.rb
```diff
@@ -4,36 +4,78 @@ module Coverband
...
+        report.each_pair do |key, line_data|
+          extended_data = {
+            'first_updated_at' => Time.now.to_i,
+            'last_updated_at' => Time.now.to_i,
+            'file_hash' => file_hash(key),
+            'data' => line_data
+          }
+          report[key] = extended_data
+        end
+        report
```
kbaum: Think you can leave off this line as each_pair returns the modified hash. Did a little test to double check:
```
irb(main):002:0> hash = {a:1, b:2}
=> {:a=>1, :b=>2}
irb(main):003:0> hash.each_pair{|k,v| hash[k] = v+1}
=> {:a=>2, :b=>3}
```

lib/coverband/adapters/base.rb
```diff
@@ -43,9 +85,26 @@ def merge_reports(new_report, old_report)
...
+        }
+      end
+
       def array_add(latest, original)
         latest.map.with_index { |v, i| (v && original[i]) ? v + original[i] : nil }
       end
+
+      def simple_report(report)
+        simple = {}
+        report.each_pair do |key, extended_data|
```
kbaum: Could use each_with_object here:

```ruby
def simple_report(report)
  report.each_with_object({}) do |(key, extended_data), simple|
    simple[key] = extended_data['data']
  end
end
```

lib/coverband/adapters/base.rb
```diff
@@ -4,36 +4,78 @@ module Coverband
...
         raise 'abstract'
       end
 
-      def save_report(_report)
-        raise 'abstract'
+      # Note: This could lead to slight race on redis
+      # where multiple processes pull the old coverage and add to it then push
+      # the Coverband 2 had the same issue,
+      # and the tradeoff has always been acceptable
+      def save_report(report)
```
kbaum: Really great that you pushed this stuff up to the base class.

PR #139, opened 2018-11-12
Coverband#start kicks off background thread

kbaum:
Unless running in rack server. In this case middleware kicks off on first request.

This PR is an addition to https://github.com/danmayer/coverband/pull/138


kbaum:
Done!

PR #141, opened 2018-11-14
Check for rails server command

kbaum:
Noticed that when starting with rails s on a rails 4 project for some
reason rack was not in the stack. This pull request looks for the rails
commands/commands_tasks server method. If it finds this method it delays
early start and waits until the rack middleware kicks off the thread on
first request.


danmayer:
looks good, +1 to merge

PR #142, opened 2018-11-15
Report data to coverband at exit

kbaum:


danmayer:
ha nice, I had considered this in the past and not done it but with the background thread, I think it is a great fit! +1 from me

kbaum:
Yeah.. i fee like it handles a quick rake task that runs in 10 seconds. Without an at_exit, we would miss that coverage.

PR #143, opened 2018-11-18
Add end to end rails5 test

kbaum:


kbaum:
Great! Started to add rails4 but didn’t get enough time on weekend.

PR #145, opened 2018-11-20
Add rails4 end to end test

kbaum:
End to end test now runs for rails5 and rails4 within travis. Using a 2nd rails4 Gemfile on travis that depends on rails4 instead of rails5.


coverband.gemspec
```diff
@@ -31,7 +31,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rake'
   spec.add_development_dependency 'redis'
   spec.add_development_dependency 'test-unit'
-  spec.add_development_dependency 'rails', '~>5'
```
danmayer: with two Gemfiles which explicitly call out Rails, do we still need or want this? @kbaum

danmayer:
added one comment and how does one run the two different Rails tests could you perhaps add notes on how to do that to the readme? @kbaum

coverband.gemspec
kbaum: Oh.. good catch! We don't need this in the gemspec anymore.

kbaum:
Will do. rails5 is the in the default Gemfile so it's just:

```
bundle exec rake test
```

For rails4:

```
BUNDLE_GEMFILE=Gemfile.rails4 bundle exec rake test
```

kbaum:
Added to README. Let me know what you think.
thx.

danmayer:
nice +1 very cool to have the end to end tests in should help with tests verifying Railsties integration work.

PR #146, opened 2018-11-22
WIP on Railtie

danmayer:
I think this might be ready.

* note I am not sure if the background reporting released in 3.0.1 was usable @kbaum see that I had to add the force report to bypass the background check... the issue I hit was that even when the background job tried to report coverage it was a no-op because the first line just returned... Perhaps I missed something?
* so far this all seems good for getting initial data when the app boots, and for the background 30-second reporting

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,9 +5,9 @@ Build Status: [![Build Status](https://travis-ci.org/danmayer/coverband.svg?bran
 <p align="center">
   <a href="#key-features">Key Features</a> •
   <a href="#coverband-demo">Coverband Demo</a> •
-  <a href="#how-to-use">How To Use</a> •
   <a href="#installation">Installation</a> •
   <a href="#configuration">Configuration</a> •
+  <a href="#how-to-use">How To Use</a> •
   <a href="#usage">Usage</a> •
   <a href="#license">License</a> •
   <a href="/changes.md">Change Log / Roadmap</a>
@@ -31,16 +31,6 @@ The primary goal of Coverband is giving deep insight into your production runtim
 ## Coverband Demo
 
 Take Coverband for a spin on the live Heroku deployed [Coverband Demo](https://coverband-demo.herokuapp.com/). The [full source code for the demo](https://github.com/danmayer/coverband_demo) is available to help with installation, configuration, and understanding of basic usage.
-
-## Example Output
-
-Since Coverband is [Simplecov](https://github.com/colszowka/simplecov) output compatible it should work with any of the `SimpleCov::Formatter`'s available. The output below is produced using the default Simplecov HTML formatter.
-
-Index Page
-![image](https://raw.github.com/danmayer/coverband/master/docs/coverband_index.png)
-
-Details on a example Sinatra app
-![image](https://raw.github.com/danmayer/coverband/master/docs/coverband_details.png)
 
 
 # Installation
@@ -155,6 +145,16 @@ Below is my Coverband workflow, which hopefully will help other best use this li
 - [Rails 5.2.x App](https://github.com/danmayer/coverage_demo)
 - [Sinatra app](https://github.com/danmayer/churn-site)
 - [Non Rack Ruby app](https://github.com/danmayer/coverband_examples)
+
+### Example Output
+
+Since Coverband is [Simplecov](https://github.com/colszowka/simplecov) output compatible it should work with any of the `SimpleCov::Formatter`'s available. The output below is produced using the default Simplecov HTML formatter.
+
+Index Page
+![image](https://raw.github.com/danmayer/coverband/master/docs/coverband_index.png)
+
+Details on a example Sinatra app
+![image](https://raw.github.com/danmayer/coverband/master/docs/coverband_details.png)
 
 ### View Coverage
 
--- a/README.md
+++ b/README.md
@@ -32,18 +32,6 @@ The primary goal of Coverband is giving deep insight into your production runtim
 
 Take Coverband for a spin on the live Heroku deployed [Coverband Demo](https://coverband-demo.herokuapp.com/). The [full source code for the demo](https://github.com/danmayer/coverband_demo) is available to help with installation, configuration, and understanding of basic usage.
 
-# How To Use
-
-Below is my Coverband workflow, which hopefully will help other best use this library.
-
-* <a href="#installation">Install Coverband</a>
-* Start your app and hit a few endpoints
-* Validate data collection and code coverage with  `rake coverband:coverage`
-* If you see app startup and recent visits showing, setup is correct
-* I generally configure the mountable web endpoint to [view the data via the web-app](https://github.com/danmayer/coverband#viewing--hosting-s3-coverband-results-in-app)
-* After Coverband has been verified to be working on production, I let it run for a few weeks.
-* Then I view the report and start submitting PRs for the team to review delete large related sets of code that no longer are in use.
-
 ## Example Output
 
 Since Coverband is [Simplecov](https://github.com/colszowka/simplecov) output compatible it should work with any of the `SimpleCov::Formatter`'s available. The output below is produced using the default Simplecov HTML formatter.
@@ -60,10 +48,6 @@ Details on a example Sinatra app
 Follow the below section to install and configure Coverband.
 
 ![coverband installation](https://raw.githubusercontent.com/danmayer/coverband/master/docs/coverband-install-resize.gif)
-
-## Prerequisites
-
-* Coverband 3.0.x+ requires Ruby 2.3+
 
 ## Gem Installation
 
@@ -151,6 +135,18 @@ run ActionController::Dispatcher.new
 * run `rake coverband:coverage` this will show app initialization coverage
 * make another request, or enough that your reporting frequency will trigger
 * run `rake coverband:coverage` and you should see coverage increasing for the endpoints you hit.
+
+# How To Use
+
+Below is my Coverband workflow, which hopefully will help other best use this library.
+
+* <a href="#installation">Install Coverband</a>
+* Start your app and hit a few endpoints
+* Validate data collection and code coverage with  `rake coverband:coverage`
+* If you see app startup and recent visits showing, setup is correct
+* I generally configure the mountable web endpoint to [view the data via the web-app](https://github.com/danmayer/coverband#viewing--hosting-s3-coverband-results-in-app)
+* After Coverband has been verified to be working on production, I let it run for a few weeks.
+* Then I view the report and start submitting PRs for the team to review delete large related sets of code that no longer are in use.
 
 # Usage
 
@@ -419,6 +415,11 @@ You could see some confusing results... To avoid this issue Coverband has a Rake
 `rake coverband:coverage_no_filters`
 
 This will build the report after disabling any `.simplecov` applied settings.
+
+# Prerequisites
+
+* Coverband 3.0.X+ requires Ruby 2.3+
+* Coverband currently requires Redis for production usage
 
 # Contributing To Coverband
 
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@ Build Status: [![Build Status](https://travis-ci.org/danmayer/coverband.svg?bran
 
 <p align="center">
   <a href="#key-features">Key Features</a> •
-  <a href="#coverband-demo">Coverband Demo</a> •
   <a href="#installation">Installation</a> •
   <a href="#configuration">Configuration</a> •
+  <a href="#coverband-demo">Coverband Demo</a> •
   <a href="#how-to-use">How To Use</a> •
   <a href="#usage">Usage</a> •
   <a href="#license">License</a> •
@@ -27,10 +27,6 @@ The primary goal of Coverband is giving deep insight into your production runtim
 * Easy to understand actionable insights from the report
 * Development mode, offers deep insight of code usage details (number of LOC execution during single request, etc) during development.
 * Mountable web interface to easily share reports
-
-## Coverband Demo
-
-Take Coverband for a spin on the live Heroku deployed [Coverband Demo](https://coverband-demo.herokuapp.com/). The [full source code for the demo](https://github.com/danmayer/coverband_demo) is available to help with installation, configuration, and understanding of basic usage.
 
 
 # Installation
@@ -77,8 +73,8 @@ After installing the gem in Rails you should have these Rake tasks
 
 ```bash
 rake -T coverband
-rake coverband:clear         # reset coverband coverage data
-rake coverband:coverage      # report runtime coverband code coverage
+rake coverband:coverage      # report runtime Coverband code coverage
+rake coverband:clear         # reset Coverband coverage data, helpful for development, debugging, etc
 ```
 
 You can view the report different ways, but the easiest is the Rake task which opens the SimpleCov formatted HTML.
@@ -135,6 +131,10 @@ end
 * make another request, or enough that your reporting frequency will trigger
 * run `rake coverband:coverage` and you should see coverage increasing for the endpoints you hit.
 
+# Coverband Demo
+
+Take Coverband for a spin on the live Heroku deployed [Coverband Demo](https://coverband-demo.herokuapp.com/). The [full source code for the demo](https://github.com/danmayer/coverband_demo) is available to help with installation, configuration, and understanding of basic usage.
+
 # How To Use
 
 Below is my Coverband workflow, which hopefully will help other best use this library.
@@ -146,8 +146,6 @@ Below is my Coverband workflow, which hopefully will help other best use this li
 * I generally configure the mountable web endpoint to [view the data via the web-app](https://github.com/danmayer/coverband#viewing--hosting-s3-coverband-results-in-app)
 * After Coverband has been verified to be working on production, I let it run for a few weeks.
 * Then I view the report and start submitting PRs for the team to review delete large related sets of code that no longer are in use.
-
-# Advanced Usage
 
 ### Example apps
 
@@ -358,12 +356,13 @@ By adding any files above you will get reporting on those files as part of your 
 
 By default Coverband has assumed you are trying to track your application code usage not all the supporting framework and library (Gems) code usage. There has been some good points and reasons folks want to track library usage, for example to find out which Gems they aren't actually using in production. See some of the discussion on [issue 21](https://github.com/danmayer/coverband/issues/21).
 
-* Using the Coverage Collector
-   * use the `safe_reload_files` feature to add the path of all gem files you wish to track
-   * --- or ---
-   * ensure you call `Coverband.start` before loading all your gems
-      * while possible this is currently hard as Rails and most environments load your whole Gemfile
-      * looking for an improve and easier way to support this.  
+How to collect gem usage with Coverband:
+
+* use the `safe_reload_files` feature to add the path of all gem files you wish to track
+* --- or ---
+* ensure you call `Coverband.start` before loading all your gems
+   * while possible this is currently hard as Rails and most environments load your whole Gemfile
+   * we are looking for an improve and easier way to support this.  
 
 
 ### Verbose Debug / Development Mode
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Build Status: [![Build Status](https://travis-ci.org/danmayer/coverband.svg?bran
 
 A gem to measure production code usage, showing a counter for the number of times each line of code that is executed. Coverband allows easy configuration to collect and report on production code usage. It reports in the background via a thread or can be used as Rack middleware, or manually configured to meet any need.
 
-Note: Coverband is not intended for test code coverage, for that we recommended using [SimpleCov](https://github.com/colszowka/simplecov).
+__Note:__ Coverband is not intended for test code coverage, for that we recommended using [SimpleCov](https://github.com/colszowka/simplecov).
 
 ## Key Features
 
@@ -35,35 +35,43 @@ Take Coverband for a spin on the live Heroku deployed [Coverband Demo](https://c
 
 # Installation
 
-Follow the below section to install and configure Coverband.
-
-![coverband installation](https://raw.githubusercontent.com/danmayer/coverband/master/docs/coverband-install-resize.gif)
+Coverband should be near zero config for Rails apps, if you have an issue with that, please [file an issue](https://github.com/danmayer/coverband/issues).
 
 ## Gem Installation
 
-Add this line to your application's Gemfile:
+Add this line to your application's `Gemfile`, remember to `bundle install` after updating:
 
 ```bash
 gem 'coverband'
 ```
 
-And then execute:
+## Rails
 
-```bash
-$ bundle
+The Railtie integration means you shouldn't need to do anything anything else.
+
+* By default Coverband will try to stored data to Redis
+	* Redis endpoint is looked for in this order: `ENV['COVERBAND_REDIS_URL']`, `ENV['REDIS_URL']`, or `localhost`
+
+## Sinatra
+
+For the best coverage you want this loaded as early as possible. I have been putting it directly in my `config.ru` but you could use an initializer, though you may end up missing some boot up coverage.
+
+```ruby
+require File.dirname(__FILE__) + '/config/environment'
+
+require 'coverband'
+Coverband.configure
+Coverband.start
+
+use Coverband::Middleware
+run ActionController::Dispatcher.new
 ```
 
-Or install it yourself as:
+# Report Generation
 
-```bash
-$ gem install coverband
-```
+There are two primary ways to generate and view the report
 
-# Configuration
-
-After you have the gem, you may need to configure additional options:
-
-### 1. Verify Rake
+## Rake Tasks
 
 After installing the gem in Rails you should have these Rake tasks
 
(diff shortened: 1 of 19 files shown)


lib/coverband/collectors/coverage.rb
```diff
@@ -31,7 +31,7 @@ def reset_instance
       end
 
       def report_coverage(force_report = false)
-        return Background.start if @background_reporting_enabled
+        return Background.start if @background_reporting_enabled && !force_report
```
kbaum: @danmayer definitely needed. I don't think I tested the actual report generation enough and was more concerned with how and when the background thread was running. Sorry about that. Should I backport your fix for a 3.02 release?

lib/coverband/utils/railtie.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module Coverband
+  class Railtie < Rails::Railtie
+    # Coverband needs to be setup before any of the initializers
+    # to capture usage of things loaded by them
+    # if one uses before_eager_load as I did previously
+    # any files that get loaded as part of railties will have no coverage
+    config.before_initialize do
+      Coverband.configure
```
kbaum: This line currently fails if there is no coverband.rb. What do you think of changing this and just going with sensible defaults?

lib/coverband/collectors/coverage.rb
kbaum: Should we consider kicking off Background.start from the middleware instead of here? Then we don't have to worry about the force_report flag.

lib/coverband/utils/railtie.rb
```diff
@@ -0,0 +1,27 @@
...
+    initializer 'coverband.configure' do |app|
+      app.middleware.use Coverband::Middleware
+    end
+
+    config.after_initialize do
+      Coverband::Collectors::Coverage.instance.report_coverage(true)
+    end
+
+    rake_tasks do
+      load 'coverband/utils/tasks.rb'
```
kbaum: Nice! I didn't realize there was a hook for rake tasks!

lib/coverband/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Coverband
-  VERSION = '3.0.1'
+  VERSION = '3.0.2.alpha'
```
kbaum: Are we targeting this for 3.0.2 or 4.x?

lib/coverband/utils/railtie.rb
```diff
@@ -0,0 +1,27 @@
...
+    # if one uses before_eager_load as I did previously
+    # any files that get loaded as part of railties will have no coverage
+    config.before_initialize do
+      Coverband.configure
+      Coverband.configuration.logger&.debug('Railtie setting up Coverband')
+      Coverband.start
+    end
+
+    initializer 'coverband.configure' do |app|
+      app.middleware.use Coverband::Middleware
```
kbaum: If we are targeting 3.0.2, then we might need to consider this line adding 2 middlewares on accident.

lib/coverband/collectors/coverage.rb
danmayer: yeah that makes sense, perhaps back to the idea of 2 distinct middlewares... one just to start background and the other for middleware reporting?

I am wondering if we could have middleware the removes itself after it kicks off background start so we don't have it adding any overhead after that starts.

lib/coverband/utils/railtie.rb
danmayer: Yeah I was thinking of trying to work towards the zero config side of things. I will do a pass and see if anything is really needed for basic running out of the gate. I am assuming for zero config we look for standard envs like REDIS_URL. We probably would want to give better errors if we can't detect anything.

lib/coverband/version.rb
danmayer: I wasn't sure yet but needed to pick something to test locally, I didn't release the alpha gem. I think to really get full benefits we will want to do some breaking changes so I am going to move it to 4.0.0

lib/coverband/collectors/coverage.rb
danmayer: I think we still need force report to bypast the reporting frequency...

I guess I wouldn't worry about a 3.0.2 release since we didn't have this enabled by default or fullly documented I don't think it is impacting anymore if we get a bug report we can point them towards this. We might get 4.0.0 done before it is an issue.

lib/coverband/utils/railtie.rb
danmayer: see latest commits, I think it would cover basic assumptions we could probably make it a bit more resilent but feels like a reasonable first pass.

lib/coverband/version.rb
danmayer: moved to 4.0.x in latest commits

lib/coverband/collectors/coverage.rb
kbaum: Ahhh.. totally didn't think about reporting frequency. Removing all of these configuration options will make the logic much easier to reason about. Does it make sense to target removing reporting_frequency altogether considering we are now reporting from the background thread?

Agreed on not releasing a 3.0.2. Let's push for 4.0.0.

lib/coverband/configuration.rb
```diff
@@ -56,6 +61,12 @@ def s3_secret_access_key
...
     def store
       return @store if @store
+      @store = if ENV['REDIS_URL']
+                 Coverband::Adapters::RedisStore.new(Redis.new(url: ENV['REDIS_URL']))
+               else
+                 Coverband::Adapters::RedisStore.new(Redis.new)
+               end
+      return @store if @store
       raise 'no valid store configured'
     end
```
kbaum: Was thinking it might make sense to try COVERBAND_REDIS_URL then REDIS_URL.

```ruby
def store
  @store ||= Coverband::Adapters::RedisStore.new(Redis.new(url: redis_url))
end

def redis_url
  ENV['COVERBAND_REDIS_URL'] || ENV['REDIS_URL']
end
```

This will give an easy way to specific an alternate redis instance in production.

danmayer: Good call adding this... changed it up a bit to make sure we pull in redis_ttl and redis_namespace as well

danmayer: ok see latest commit I think this is resolved

lib/coverband/collectors/coverage.rb
danmayer: ok take a look at latest commits @kbaum I refectored this a bit and moved background start to the middleware

danmayer:
updated to get all tests passing

lib/coverband/configuration.rb
```diff
@@ -55,19 +60,38 @@ def s3_secret_access_key
...
       elsif store.is_a?(String)
         @store = Coverband::Adapters::FileStore.new(store)
       end
     end
+
+    private
+
+    def coverband_redis
+      # Redis doesn't default to correct local host if you pass url: nil
+      redis_url ? Redis.new(url: redis_url) : Redis.new
```
kbaum: Not that important but Redis.new(url: nil) seems to work for me:
```
[4] pry(main)> Redis.new(url: nil)
#<Redis client v4.0.3 for redis://127.0.0.1:6379/0>
[5] pry(main)>
```

lib/coverband/configuration.rb
```diff
@@ -55,19 +60,38 @@ def s3_secret_access_key
...
+    def redis_url
+      ENV['COVERBAND_REDIS_URL'] || ENV['REDIS_URL']
+    end
+
+    def redis_store_options
+      { ttl: Coverband.configuration.redis_ttl,
+        redis_namespace: Coverband.configuration.redis_namespace }
+    end
+
+    def redis_options
```
kbaum: Are both redis_options and redis_store_options methods needed?

lib/coverband/integrations/middleware.rb
```diff
@@ -9,7 +9,11 @@ def initialize(app)
     def call(env)
       @app.call(env)
     ensure
-      Coverband::Collectors::Coverage.instance.report_coverage
+      if Coverband.configuration.background_reporting_enabled
+        Coverband::Background.start
+      else
+        Coverband::Collectors::Coverage.instance.report_coverage
```
kbaum: Do we even want to keep this option around?

lib/coverband/configuration.rb
danmayer: weird I swear I had had an issue with that in the past...

lib/coverband/configuration.rb
danmayer: oops no I meat to get rid of the redis_options doing that now...

lib/coverband/configuration.rb
danmayer: fixed

lib/coverband/integrations/middleware.rb
danmayer: yeah see other comments, I'd rather not get rid of the middleware flow until we ensure on all setups the background thread works well. Feels like it would be good to have both as an option until we get a bit more real-world testing and feedback on the new background flow.

kbaum: Agreed.

README.md
```diff
@@ -96,86 +87,48 @@ $ gem install coverband
...
-After you have the gem, you will want to setup the needed integration:
+After you have the gem, you may need to configure additional options:
 
-1. Coverband Config File Setup
-2. Configure Rake
-3. Insert Coverband Rack Middleware
+### 1. Verify Rake
 
-### 1. Coverband Config File Setup
+After installing the gem in Rails you should have these Rake tasks
```
kbaum: I think it's important that the readme gives a feeling of a drop in gem.. especially for rails. Was thinking the basic installation would be communicated near the top. Those who are interested in additional configuration options have to do a bit more digging in. Something like:

## General installation
### Rails
### Sinatra
## Report Generation
### Rake Task
### Viewing the Report in App
#### Rails
#### Sinatra
## Additional Configuration

kbaum: Just wanted to add.. happy to take a crack at it if you like the basic idea.

lib/coverband/configuration.rb
```diff
@@ -55,19 +59,28 @@ def s3_secret_access_key
     end
 
     def store
-      return @store if @store
-      raise 'no valid store configured'
+      @store ||= Coverband::Adapters::RedisStore.new(Redis.new(url: redis_url), redis_store_options)
     end
 
     def store=(store)
       if store.is_a?(Coverband::Adapters::Base)
```
kbaum: What do you think of removing the option of passing in a redis instance here?  Make this config option only take a store. Would really clean up this method and it seems pretty straight forward.

danmayer: yeah that was for old backward compatibility now is a good time to remove. Doing that. Thanks

lib/coverband/configuration.rb
```diff
@@ -55,19 +59,28 @@ def s3_secret_access_key
...
       end
     end
+
+    private
+
+    def redis_url
+      ENV['COVERBAND_REDIS_URL'] || ENV['REDIS_URL']
+    end
+
+    def redis_store_options
```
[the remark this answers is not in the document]
danmayer: Not having namespace or ttl as part of the internal knowledge of the system would be nice, but would be very different and remove some specific features. Naming sharing a single Redis for multiple environments or applications...

Basically, redis-rb doesn't support ttl or namespace at a connection level at all. https://github.com/redis/redis-rb

TTL you simply have to call per key which is what Coverband does unless I am missing something https://github.com/danmayer/coverband/blob/feature/railtie/lib/coverband/adapters/redis_store.rb#L32

---

Namespace... It is kind of an annoyingly controversial issue...

* https://github.com/redis/redis-rb/issues/222

This caused various folks to declare the redis-namespace gem a hack, "which should never be used". While others want to use a single Redis without polluting the top level namespace... https://www.mikeperham.com/2017/04/10/migrating-from-redis-namespace/

Basically, I don't want to have to add in the Redis namespace gem... and what Coverband does it also a simple naming hack... https://github.com/danmayer/coverband/blob/feature/railtie/lib/coverband/adapters/redis_store.rb#L27

Reasons to keep it, one can have staging(s) and production share a Redis using this feature, which is the original reason it was requested in the past. Reasons to get rid of it... is that it is just a naming convention, we could push folks asking them to Redis dbs as Sidekiq did, but seems it has performance issues.

I think it is worth supporting multiple apps and environments via these options.

lib/coverband/configuration.rb
danmayer: done

lib/coverband/configuration.rb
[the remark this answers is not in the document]
kbaum: Totally makes sense. Thanks for the explanation! Guess I’m just giving all the configuration options a second look while we have the chance.

README.md
danmayer: Thanks, I took another pass moving some of the more advanced stuff to the end and trying to simplify the earlier parts of the Readme. Take a look and let me know if that looks better @kbaum

README.md
```diff
@@ -5,190 +5,126 @@ Build Status: [![Build Status](https://travis-ci.org/danmayer/coverband.svg?bran
...
+  # ... lots of routes
+  mount Coverband::Reporters::Web.new, at: '/coverage'
 end
 ```
 
-#### For Sinatra apps
+__NOTE__: ADD PASSWORD PROTECTION OR YOU CAN EXPOSE ALL YOUR SOURCE CODE
 
-For the best coverage you want this loaded as early as possible. I have been putting it directly in my `config.ru` but you could use an initializer, though you may end up missing some boot up coverage.
+It is easy to add some basic protect around the coverage data, below shows how you can use devise or basic auth, by adding a bit of code to your `config/routes.rb` file.
```
kbaum: "basic protection"

README.md
```diff
@@ -199,97 +135,107 @@ run ActionController::Dispatcher.new
...
 
-### Example apps
+ Below is an example config file for a Rails 5 app:
 
-- [Rails 5.2.x App](https://github.com/danmayer/coverage_demo)
-- [Sinatra app](https://github.com/danmayer/churn-site)
-- [Non Rack Ruby app](https://github.com/danmayer/coverband_examples)
+```ruby
+#config/coverband.rb
+Coverband.configure do |config|
```
kbaum: Think we should mention that we can set the environment variables or configure through coverband with config.s3_access_key_id and config.s3_secret_access_key?

README.md
```diff
@@ -305,6 +251,7 @@ def after_perform(*args)
 end
 ```
 
+(no need to do this for Rails, but a Rack app maybe)
```
kbaum: This part may be confusing. For sidekiq and no rails, the default background thread removes the need for custom sidekiq middleware. The only custom code needed should be Coverband.configure and Coverband.start. I am thinking we should probably even remove the need for this. Maybe:

```ruby
if defined?(Sidekiq) && !defined?(Rails)
  Sidekiq.configure_server do
    Coverband.configure
     Coverband.start
  end
end
```

While we are on the topic of configure and start, would it be possible to just call these methods on require of coverband.rb and remove this requirement altogether?

README.md
danmayer: fixed

README.md
danmayer: good call updated docs around this as well

README.md
danmayer: Yeah so I was thinking for Ruby but not Rails Sidekiq apps... I think your idea of automatically calling configure and start even outside of the railtie is an interesting one. Let me run some tests on some plain ruby scripts and see if I have any Sinatra apps to test with, It seems like a good idea, but I would want to test it out a bit on some samples first.

README.md
```diff
@@ -337,6 +284,8 @@ Coverband::Collectors::Coverage.instance.report_coverage
 
 ### Manual Configuration (for cron jobs / Raketasks)
```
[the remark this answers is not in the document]
danmayer: Same as above, I am going to test this a bit if it works well I will drop these doc sections and replace with an explanation.

README.md
danmayer: ok, take a look at the proof of concept of this idea...

added on latest commit to this branch... https://github.com/danmayer/coverband/commit/3240a161f8739150ed18473647284914fffaaf97

An example script which collects its own runtime data via this method
https://github.com/danmayer/coverband_demo/pull/2/files#diff-7ce9c6873a144d3d0eb1ca7010c03a86R1

The Coverband demo Rails app still works as well.

danmayer: Let me know if you see any issues with this, if not I will try to do a bit more testing and clean up docs @kbaum

README.md
```diff
@@ -5,190 +5,126 @@ Build Status: [![Build Status](https://travis-ci.org/danmayer/coverband.svg?bran
...
 * Low performance overhead
-* Various controls from sampling, data collection rate, etc to further control performance
-* Ignore directories to avoid overhead data collection on vendor, lib, etc.
+* Very simple setup and configuration
+* Out of the box support for all standard code execution paths (web, cron, background jobs, rake tasks, etc)
+* Easy to understand actionable insights from the report
 * Development mode, offers deep insight of code usage details (number of LOC execution during single request, etc) during development.
-* Easy setup for any Ruby Rack based web framework (Rails, Sinatra, etc)
-* Allows for integration with any other Ruby application flows (background jobs, crons, scripts)
+* Mountable web interface to easily share reports
```
kbaum: The features section looks great!

README.md
```diff
@@ -5,190 +5,126 @@ Build Status: [![Build Status](https://travis-ci.org/danmayer/coverband.svg?bran
...
-## Example Output
-
-Since Coverband is [Simplecov](https://github.com/colszowka/simplecov) output compatible it should work with any of the `SimpleCov::Formatter`'s available. The output below is produced using the default Simplecov HTML formatter.
-
-Index Page
-![image](https://raw.github.com/danmayer/coverband/master/docs/coverband_index.png)
-
-Details on a example Sinatra app
-![image](https://raw.github.com/danmayer/coverband/master/docs/coverband_details.png)
+Take Coverband for a spin on the live Heroku deployed [Coverband Demo](https://coverband-demo.herokuapp.com/). The [full source code for the demo](https://github.com/danmayer/coverband_demo) is available to help with installation, configuration, and understanding of basic usage.
```
kbaum: I think the demo section should be moved lower. When i look at a readme, i want to see what the gem is and then how to install it.

README.md
```diff
@@ -5,190 +5,126 @@ Build Status: [![Build Status](https://travis-ci.org/danmayer/coverband.svg?bran
...
 ## Gem Installation
 
-Add this line to your application's Gemfile:
+Add this line to your application's `Gemfile`, remember to `bundle install` after updating:
 
 ```bash
 gem 'coverband'
 ```
 
-And then execute:
```
kbaum: Love deleting this part.

README.md
```diff
@@ -199,97 +135,116 @@ run ActionController::Dispatcher.new
...
+* <a href="#installation">Install Coverband</a>
+* Start your app and hit a few endpoints
+* Validate data collection and code coverage with  `rake coverband:coverage`
+* If you see app startup and recent visits showing, setup is correct
+* I generally configure the mountable web endpoint to [view the data via the web-app](https://github.com/danmayer/coverband#viewing--hosting-s3-coverband-results-in-app)
+* After Coverband has been verified to be working on production, I let it run for a few weeks.
+* Then I view the report and start submitting PRs for the team to review delete large related sets of code that no longer are in use.
 
-bundle install
+# Advanced Usage
```
kbaum: I think we can delete this header?

README.md
```diff
@@ -400,9 +358,6 @@ By adding any files above you will get reporting on those files as part of your
 
 By default Coverband has assumed you are trying to track your application code usage not all the supporting framework and library (Gems) code usage. There has been some good points and reasons folks want to track library usage, for example to find out which Gems they aren't actually using in production. See some of the discussion on [issue 21](https://github.com/danmayer/coverband/issues/21).
 
-* Using the Tracepoint Collector
-   * just make sure to set `Coverband.configuration.include_gems = true` in your config options
-   * note by tracking all the additional usage this can have significant performance impacts of the Tracepoint implementation
 * Using the Coverage Collector
```
kbaum: Since there is only 1 kind of collector now, we might want to reword this to be more informational. Like, "How it works" or something like that. Perhaps also moving to a wiki page?

kbaum:
The README is really looking great. If we remove the need for calling configure and start, it will be even simpler

RE: automatically calling `Coverband.configure` and `Coverband.start` on require of the gem. Can't think of any downside to calling configure early. Coverband.start will cause the coverage to be started much earlier and possibly before the require of many other gems. This means that ruby Coverage will be tracking more lines of code. Not sure how much this impacts performance if at all but worth thinking about.

README.md
danmayer: moved down.

README.md
```diff
@@ -5,190 +5,126 @@ Build Status: [![Build Status](https://travis-ci.org/danmayer/coverband.svg?bran
...
-require 'coverband/utils/tasks'
-```
-This should give you access to a number of Coverband tasks
+## Rake Tasks
+
+After installing the gem in Rails you should have these Rake tasks
 
 ```bash
 rake -T coverband
 rake coverband:clear         # reset coverband coverage data
```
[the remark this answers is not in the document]
danmayer: I updated the comment, but I think it is helpful for folks when trying to setup or do development

` reset Coverband coverage data, helpful for development, debugging, etc`

README.md
danmayer: done

README.md
danmayer: updated and cleaned up a bit. Yeah, I do think some of these might be good to move to the Wiki page. I think I will do that outside of this PR as it is getting a bit big ;)

danmayer:
Yeah the only downside is we would, in theory, start the collector much earlier during code loading... so Coverage will be collecting data on many more Gems and Rails internals itself, while the Railtie would mostly just collect for App code... I think I will do some perf testing now to see if it seems significant enough care about, I am hoping not.

README.md
danmayer: removed all these docs as we have autostart on require working now

README.md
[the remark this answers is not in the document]
danmayer: removed

danmayer:
screen shot 2018-11-28 at 8 27 05 pm

Railtie benchmark on the left and the Autostart on the right... mean request time was exactly the same for both methods

danmayer:
Alright, anything else we think should be in 4.0.0 release @kbaum or just some more perf checks and alpha testing.

lib/coverband/collectors/coverage.rb
```diff
@@ -17,27 +17,25 @@ def self.instance
...
         self
       end
 
       def report_coverage(force_report = false)
-        return Background.start if @background_reporting_enabled
         return if !ready_to_report? && !force_report
-
         unless @store
           @logger.debug 'no store set, no-op'
           return
```
kbaum: Do we want to raise an error here? Could be confusing why nothing is working. Another approach could be to throw an error on Coverband.configure. That way we never even get here.

lib/coverband/collectors/coverage.rb
```diff
@@ -74,14 +71,12 @@ def ready_to_report?
       end
 
       def get_new_coverage_results
-        coverage_results = nil
-        @semaphore.synchronize { coverage_results = new_coverage(::Coverage.peek_result.dup) }
-        coverage_results
+        @semaphore.synchronize { new_coverage(::Coverage.peek_result.dup) }
```
kbaum: Minor note but when we are using background thread, we should be able to avoid the `@semaphore` since there should always be 1 thread.

lib/coverband/collectors/coverage.rb
kbaum: One other thought, is a nil @store even possible? 
https://github.com/danmayer/coverband/pull/146/files#diff-1f4b05f150269d04438736256c27b0a8R62

kbaum:
Let's 🚢 alpha!

One other note @dnasseri and I started integrating 4.0 into our project yesterday and it is looking good so far. Probably release it into our app today or tomorrow.

lib/coverband/collectors/coverage.rb
danmayer: yep thought about it but since I want to keep middleware option around for now we need to keep it

lib/coverband/collectors/coverage.rb
danmayer: Yeah I had thought a bit about that... I guess it would be nice to push this into configure... basically if you don't set a store and the default one fails for some reason raise an error. 

The issue would be if the error only happens in prod it could be a scary experience... I have tried to make sure Coverband is always safe and doesn't hurt prod when it isn't working... Thoughts?

kbaum: Looks like we are already rescuing all errors anyway?

https://github.com/danmayer/coverband/pull/146/files#diff-181f764c69748a3ceb0cc64e4a37ebaeR43

danmayer: yeah true... I change to a raise which gets caught and logged, I don't think configure would get caught since that would happen outside of this rescue. Anyways should make it a bit more obvious something is wrong.

danmayer:
OK 4.0.0.alpha is live for testing https://coverband-demo.herokuapp.com/coverage?

lib/coverband/collectors/coverage.rb
```diff
@@ -31,10 +31,7 @@ def reset_instance
 
       def report_coverage(force_report = false)
         return if !ready_to_report? && !force_report
-        unless @store
-          @logger.debug 'no store set, no-op'
-          return
-        end
+        raise 'no Coverband store set' unless @store
```
kbaum: This is great because it will make it so much easier to debug people's issues.

danmayer:
I have 4.0.0 running on a real app on staging and will get it to production tomorrow

danmayer:
All has worked well for me @kbaum think we are good to merge and release? Or anything else we should check?

dnasseri:
@kbaum @danmayer I've been running 4.0.0.alpha in production for two days now and everything is looking great! Nice work to both of you!

kbaum:
@danmayer - let's merge and release!

danmayer:
OK 4.0.0 is live!

PR #149, opened 2018-12-15
Fix test race condition

kbaum:
When running this test alone, it would fail. 

```
➜  coverband git:(master) bundle exec m test/unit/rails_full_stack_test.rb
Run options: -n "/^(test_this_is_how_we_do_it|test_memory_usage)$/" --seed 28906

# Running:

.F

Failure:
RailsFullStackTest#test_this_is_how_we_do_it [/Users/karlbaum/workspace/coverband/test/unit/rails_full_stack_test.rb:14]:
Expected: [1, 1, 1, nil, nil]
  Actual: nil
```

When running with rake test it would pass. I am not sure why it ever passed honestly as it was
depending on a background thread reporting to redis every 30 seconds.

### DIFF
--- a/lib/coverband.rb
+++ b/lib/coverband.rb
@@ -29,8 +29,7 @@ module Coverband
   end
 
   def self.configure(file = nil)
-    configuration_file = file || CONFIG_FILE
-
+    configuration_file = file || ENV.fetch('COVERBAND_CONFIG', CONFIG_FILE)
     configuration
     if block_given?
       yield(configuration)
--- a/lib/coverband/integrations/background.rb
+++ b/lib/coverband/integrations/background.rb
@@ -4,6 +4,15 @@ module Coverband
   class Background
     @semaphore = Mutex.new
 
+    def self.stop
+      @semaphore.synchronize do
+        if @thread
+          @thread.exit
+          @background_reporting_running = false
+        end
+      end
+    end
+
     def self.start
       return if @background_reporting_running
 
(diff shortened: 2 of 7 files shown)


kbaum:
I really like that sleep idea! It’s a much more real world test. I’ll give that a shot.

kbaum:
The sleep approach seems to work well. The problem I ran into was that it turned out that the rails dummy app coverband.rb was not being read. To address that problem I had to add an environment variable. At the end of the test I reset that environment variable and I also stopped the background thread. I think our tests easily leak global state. I noticed that the background test started to fail and I needed to explicitly reset the background_reporting_sleep_seconds:

https://github.com/danmayer/coverband/pull/149/files#diff-96b6bd6385978b0f49338847dc76e4f5R11

Maybe we need a simple way to reset everything between tests?

danmayer:
hmm yeah so that was the original purpose of coverage#reset_instance 

https://github.com/danmayer/coverband/blob/master/lib/coverband/collectors/coverage.rb#L17

I guess perhaps we need something that resets at a higher level or configuration#reset or something. @kbaum

danmayer:
oh, I do something like this in some tests... @kbaum perhaps we make that a bit more generic and move it to the test helper in a before_each type block.

https://github.com/danmayer/coverband/blob/master/test/unit/full_stack_test.rb#L10

kbaum:
I think a higher level reset makes sense considering the configuration is a singleton and carries state between tests. We now also need to think about the background thread running.

kbaum:
Created a PR for resetting after each test in test_helper.

https://github.com/danmayer/coverband/pull/151

danmayer:
nice looks good I merged #151 so this one can likely move forward any other changes on this PR or is it ready to go now

kbaum:
This one looks ready to go. Thx.

danmayer:
nice

PR #150, opened 2018-12-15
Add end to end feature test

kbaum:
Adds a capybara feature test which should verify end to end workflow for rails5 and rails4.

This PR includes this: 
https://github.com/danmayer/coverband/pull/149

Which is off of master so that is merged in as well.

This PR is requesting merge into this:
https://github.com/danmayer/coverband/pull/148

Sorry for the git confusion. Let me know if that all makes sense.

### DIFF
--- a/lib/coverband.rb
+++ b/lib/coverband.rb
@@ -39,7 +39,7 @@ module Coverband
     if block_given?
       yield(configuration)
     elsif File.exist?(configuration_file)
-      require configuration_file
+      load configuration_file
     else
       configuration.logger&.debug('using default configuration')
     end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -15,12 +15,18 @@ SimpleCov.start do
   add_filter '/config/'
 end
 
-class Minitest::Test
-  def teardown
+module Coverband::Test
+  def self.reset
     Coverband.configuration.store.clear!
     Coverband.configuration.reset
     Coverband::Collectors::Coverage.instance.reset_instance
     Coverband::Background.stop
+  end
+end
+
+class Minitest::Test
+  def setup
+    Coverband::Test.reset
   end
 end
 
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -3,7 +3,7 @@
 require 'rubygems'
 require 'simplecov'
 require 'minitest/autorun'
-require 'mocha/setup'
+require 'mocha/minitest'
 require 'ostruct'
 require 'json'
 require 'redis'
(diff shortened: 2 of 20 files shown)


danmayer:
looks good... it does add a number of bigger dev dependencies (nokogiri looking at you). I guess how much do you think we benefit from driving with capybara vs. using the rack test helpers... Like we did on the web endpoint tests...

https://github.com/danmayer/coverband/blob/master/test/unit/reports_web_test.rb#L33

basically, it would still do your full test flow I think but we would have to trigger the requests vs use the click link helper.... then have to pull out the 100% check a bit more manually since we can't parse html.

I am happy enough to add the dependencies if you think it is worth it. As we add more custom HTML and better templating I could see the new dependencies being more valuable for the additional web flows.

Let me know @kbaum

kbaum:
Not sure. Bringing in capybara definitely comes with a lot of dependencies. On the other hand the dsl is kind of nice and we will most likely be adding more feature style tests. We could even add another test today validating the code is being displayed properly. Is there much of a downside to adding dev dependencies? Let me know what you think.

lib/coverband.rb
```diff
@@ -34,13 +34,12 @@ class << self
...
   def self.configure(file = nil)
-    configuration_file = file || CONFIG_FILE
-
+    configuration_file = file || ENV.fetch('COVERBAND_CONFIG', CONFIG_FILE)
     configuration
     if block_given?
       yield(configuration)
     elsif File.exist?(configuration_file)
-      require configuration_file
+      load configuration_file
```
kbaum: @danmayer I had to change this to load because when running all the tests at once, the configuration file may have already been loaded. If it had already been loaded, this method because essentially a noop and the configuration was not set again. This is a problem in an environment where we are running multiple tests scenarios with different configurations.

test/test_helper.rb
```diff
@@ -15,6 +15,21 @@
   add_filter '/config/'
 end
 
+module Coverband::Test
+  def self.reset
```
kbaum: @danmayer pulled this into a separate class since rails full stack test does not seem to hit MinitTest::Test#setup.

lib/coverband.rb
danmayer: makes sense

test/unit/rails_full_stack_test.rb
```diff
@@ -2,17 +2,36 @@
...
 class RailsFullStackTest < ActionDispatch::IntegrationTest
+  include Capybara::DSL
+  include Capybara::Minitest::Assertions
+
   def setup
-    Coverband.configuration.store.clear!
+    Coverband::Test.reset
+    Capybara.default_driver = :selenium_chrome_headless
+    Capybara.server = :webrick
+    Coverband.configure("#{Rails.root}/config/coverband.rb")
```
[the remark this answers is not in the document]
danmayer: hmm yeah I guess it is a bit hard to allow a custom config location and configure on boot... we could allow customer config location via an ENV var.

kbaum:
@danmayer - I think this is ready for another look.

PR #151, opened 2018-12-18
Reset coverband config, instance, and background thread

kbaum:
After each test, clean up reset coverband global state to avoid test pollution.


danmayer:
nice looks good +1

Issue #158, opened 2018-12-30
Coverage data since...


kbaum:
The longer code has gone without being hit, the more confident we are to delete it. Let's add the coverage start date to each file in our coverage report. Something like?

![image](https://user-images.githubusercontent.com/96786/50548444-84e82480-0c1a-11e9-933b-3cc2a0d0d0fd.png)

danmayer:
yep we have the data and now own the full rendering... I guess should we add this in before 4.0.1 as it is a pretty cool new feature to add to the list? @kbaum

danmayer:
I might be able to knock it out today

danmayer:
ok take a look https://github.com/danmayer/coverband/pull/160

danmayer:
live in release 4.0.1

PR #160, opened 2018-12-30
add support for coverage time closing #158

danmayer:
This shows when a file was first seen and when it last had any coverage activity.

screen shot 2018-12-30 at 9 51 18 am

Some further cleanup is now able to be done but this seems good and should mean we can get it in this release ;)

### DIFF
--- a/lib/coverband/adapters/base.rb
+++ b/lib/coverband/adapters/base.rb
@@ -31,7 +31,7 @@ module Coverband
 
       # TODO: deprecate / remove?
       def covered_lines_for_file(file)
-        coverage[file] ? (coverage[file]['data'] || []) : []
+        Array(coverage.dig(file, 'data'))
       end
 
       protected
--- a/lib/coverband/adapters/base.rb
+++ b/lib/coverband/adapters/base.rb
@@ -22,15 +22,16 @@ module Coverband
       end
 
       def coverage
-        simple_report(get_report)
+        get_report
       end
 
       def covered_files
         coverage.keys || []
       end
 
+      # TODO: deprecate / remove?
       def covered_lines_for_file(file)
-        coverage[file] || []
+        coverage[file] ? (coverage[file]['data'] || []) : []
       end
 
       protected
--- a/lib/coverband/utils/source_file.rb
+++ b/lib/coverband/utils/source_file.rb
@@ -81,9 +81,27 @@ module Coverband
       # The array of coverage data received from the Coverage.result
       attr_reader :coverage
 
-      def initialize(filename, coverage)
+      # the date this version of the file first started to record coverage
+      attr_reader :first_updated_at
+      # the date this version of the file last saw any coverage activity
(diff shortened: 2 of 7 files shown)


lib/coverband/adapters/base.rb
```diff
@@ -22,15 +22,16 @@ def save_report(report)
...
       end
 
       def covered_files
         coverage.keys || []
       end
 
+      # TODO: deprecate / remove?
       def covered_lines_for_file(file)
-        coverage[file] || []
+        coverage[file] ? (coverage[file]['data'] || []) : []
```
kbaum: Could use Hash#dig here?
```ruby
Array(coverage.dig(file, data))
```

lib/coverband/utils/source_file.rb
```diff
@@ -81,9 +81,27 @@ def status
...
+      # the date this version of the file first started to record coverage
+      attr_reader :first_updated_at
+      # the date this version of the file last saw any coverage activity
+      attr_reader :last_updated_at
+      NOT_AVAILABLE = 'not available'
+
+      def initialize(filename, file_data)
         @filename = filename
-        @coverage = coverage
+        if file_data.is_a?(Hash)
```
kbaum: Is this needed for backwards compatibility?

lib/coverband/utils/source_file.rb
```diff
@@ -81,9 +81,27 @@ def status
...
+          @coverage = file_data['data']
+          begin
```
kbaum: Avoid the error handing?
```ruby
@first_updated_at = @last_updated_at = NOT_AVAILABLE
@first_updated_at = Time.at(file_data['first_updated_at']) if file_data['first_updated_at']
@last_updated_at =  Time.at(file_data['last_updated_at']) if file_data['last_updated_at']
```

lib/coverband/utils/source_file.rb
danmayer: This is needed for files that aren't in the coverage data and need to get reported at `0%`

lib/coverband/utils/source_file.rb
danmayer: nice switched

lib/coverband/adapters/base.rb
danmayer: nice updated

lib/coverband/utils/source_file.rb
kbaum: Oh gotcha. I guess we could eventually simplify this by just having everything default to a hash?

danmayer: yeah, I think if I do some further refactoring we can change remove these checks... and we can remove that other line I added a todo to remove... but it will required a bit more of a rewrite than I want to do for fast-tracking it into the release ;)

kbaum: 👍

PR #165, opened 2019-01-13
Update README for 4.0.1

kbaum:


README.md
```diff
@@ -28,10 +29,15 @@ The primary goal of Coverband is giving deep insight into your production runtim
 * Development mode, offers deep insight of code usage details (number of LOC execution during single request, etc) during development.
 * Mountable web interface to easily share reports
 
-
 # Installation
 
-Coverband should be near zero config for Rails apps, if you have an issue with that, please [file an issue](https://github.com/danmayer/coverband/issues).
+## Redis
```
kbaum: Thought we could use mention of redis a bit earlier since it is required.

README.md
```diff
@@ -132,18 +104,6 @@ end
 
 Take Coverband for a spin on the live Heroku deployed [Coverband Demo](https://coverband-demo.herokuapp.com/). The [full source code for the demo](https://github.com/danmayer/coverband_demo) is available to help with installation, configuration, and understanding of basic usage.
 
-# How To Use
```
kbaum: Feel like this section might be not needed anymore as it is covered elsewhere.

README.md
```diff
@@ -185,6 +145,8 @@ You may need to configure Coverband you can either do that passing in all config
 ```ruby
 #config/coverband.rb
+  config.store = Coverband::Adapters::RedisStore.new(Redis.new(url: ENV['MY_REDIS_URL']))
```
kbaum: Added some more typical configuration settings since now s3 is something that will most often not be needed.

README.md
```diff
@@ -60,65 +64,33 @@ use Coverband::Middleware
...
-```
-
-You can view the report different ways, but the easiest is the Rake task which opens the SimpleCov formatted HTML.
-
-`rake coverband:coverage`
-
-This should auto-open in your browser, but if it doesn't the output file should be in `coverage/index.html`
-
-## Viewing the Report in App
+# Coverage Report
```
kbaum: Tried to simplify the code example here.

README.md
danmayer: makes sense

README.md
danmayer: makes sense

PR #172, opened 2019-01-20
Remove simplecov section from readme

kbaum:
Don't think this section applies anymore?

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,31 +260,6 @@ If you are trying to debug locally wondering what code is being run during a req
       ...
      [517, 1617], [516, 38577]]
 
-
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -260,31 +260,6 @@ If you are trying to debug locally wondering what code is being run during a req
...
-  add_filter '/spec/'
-  add_filter '/config/'
-  add_filter '/vendor/'
-  add_filter 'userevents'
-end
-```
-
-You could see some confusing results... To avoid this issue Coverband has a Rake task that will ignore all Simplecov filters.
-
-`rake coverband:coverage_no_filters`
```
kbaum: Should I also remove this rake task as well? I noticed it doesn't work.

PR #177, opened 2019-01-26
Add resque support

kbaum:
Out of box support for resque. Current implementation will report to redis for every job execution.

### DIFF
--- a/lib/coverband/integrations/resque.rb
+++ b/lib/coverband/integrations/resque.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+Resque.after_fork do |job|
+  Coverband.start
+end
+
+module Coverband
+  module ResqueWorker
+    def perform
+      super
+    ensure
+      Coverband::Collectors::Coverage.instance.report_coverage(true)
+    end
+  end
+end
+
+Resque::Job.prepend(Coverband::ResqueWorker)
+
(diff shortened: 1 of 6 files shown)


lib/coverband/integrations/resque.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+Resque.after_fork do |job|
+  Coverband.start
+end
+
+module Coverband
```
danmayer: nice this is very clean, well done

danmayer:
merge when ever, I am not concerned on the -0.2% test change

kbaum:
Thanks! Do you think we can ask the company that was having the memory issues with reque to try this out?

kbaum:
I noticed the resque test is sporadically failing.

https://github.com/danmayer/coverband/issues/178

I'll try to figure out that issue and then do an alpha release?

Thanks for the invite to coverband_demo!

danmayer:
cool makes sense to try to fix test if we can first... I am working on another feature, but may or may not make 4.0.2 release. We will see how long it takes to get some more feedback about resque and if we can reproduce the memory lead.

I am working on adding Gem coverage support via an optional flag. Reported as a new tab on the UI that lists all gems loaded and basic percentage usage per gem.

kbaum:
Sounds good. That sounds like an amazing feature! So many gems that go
unused in big rails projects.

PR #181, opened 2019-01-29
Fix sporadic failure of resque test

kbaum:
Was difficult to figure this out as I was only able to reproduce the problem sporadically on travis. Theory is that the background job thread running on parent process was overwriting what the child process was writing to redis. This is kind of a known race condition in coverband that two different processes can compete with each other.

Have run the tests over and over again on travis and haven't had an error so i think we got it.

### DIFF
--- a/lib/coverband/integrations/resque.rb
+++ b/lib/coverband/integrations/resque.rb
@@ -9,7 +9,6 @@ module Coverband
     def perform
       super
     ensure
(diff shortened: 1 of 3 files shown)


kbaum:
Going to merge this one since it fixes build and only touches tests.

kbaum:
Fixes #181

danmayer:
cool thanks

PR #183, opened 2019-01-30
WIP: Feature/gem tracking

danmayer:
OK, I still need to do a bit more cleanup, and fill out the test coverage on the new gem functionality. I have tested this on two apps and it seems to work. When displaying all of the source files the report is very slow... with the default option hiding gem details so you only see the overall gem usage it seems acceptable even on a large app, still a bit slow, and likely worth improving in the long run, but probably OK as the pilot release of the gem functionality.

I still need to do some performance regression tests to ensure this doesn't slow down code with the gem tracking feature disabled.

My thoughts is the first release that includes this all the gem features are off by default and only optionally enabled. That is likely the best default for the gem feature in general.


danmayer:
OK, @kbaum I guess ready for some initial feedback, but I still have some work to do before this is really ready to merge.

lib/coverband/reporters/base.rb
```diff
@@ -61,16 +62,31 @@ def merge_arrays(first, second)
...
+          # filename = filename.gsub('./', roots.last)
+          # above only works for app files
+          # we need to rethink some of this logic
+          # gems aren't at project root and can have multiple locations
+          roots.each do |root|
+            if File.exist?(relative_filename.gsub('./', root))
+              local_filename = relative_filename.gsub('./', root)
+              break
+            end
+          end
```
kbaum: Not sure if this reads better using find to return the first root that has the file?
```ruby
  root = roots.find { |root| File.exist?(relative_filename.gsub('./', root)) }
  relative_filename.gsub('./', root)
```

lib/coverband/utils/gem_list.rb
```diff
@@ -0,0 +1,46 @@
...
+
+####
+# An array of  FileLists instances with helpers to roll up the stats
+# methods for calculating coverage across them etc.
+####
+module Coverband
+  module Utils
+    class GemList < FileList
+      # Returns the count of lines that have coverage
+      def covered_lines
```
kbaum: Using `to_a` on these methods instead of the empty check might work? 
```ruby
def covered_lines
  to_a.map(&:covered_lines).sum
end
```
Looks like calling `to_a` on an exiting array, just returns the same instance:
```
irb(main):012:0> a = [:test]
=> [:test]
irb(main):013:0> a.object_id == a.to_a.object_id
=> true
irb(main):014:0>
```
And `nil.to_a` returns `[]`.

lib/coverband/utils/result.rb
```diff
@@ -100,6 +100,38 @@ def coverage
...
+      def filter_to_groups(files)
+        grouped = {}
+        grouped_files = []
+        grouped_gems = {}
+        gem_lists = []
+        Coverband.configuration.groups.each do |name, filter|
+          if name == 'Gems'
+            gems_files = files.select { |source_file| source_file.filename =~ /#{filter}/ }
+            gems_files.each do |source_file|
+              grouped_gems[source_file.gem_name] = [] unless grouped_gems[source_file.gem_name]
```
kbaum:
```ruby
grouped_gems[source_file.gem_name] ||= []
```

lib/coverband/utils/result.rb
```diff
@@ -100,6 +100,38 @@ def coverage
...
+      # Applies the configured groups to the given array of SimpleCov::SourceFile items
+      #
+      def filter_to_groups(files)
+        grouped = {}
+        grouped_files = []
+        grouped_gems = {}
+        gem_lists = []
+        Coverband.configuration.groups.each do |name, filter|
+          if name == 'Gems'
+            gems_files = files.select { |source_file| source_file.filename =~ /#{filter}/ }
```
kbaum: Could this possibly be replaced with `group_by`?

```ruby
grouped_gems = files.select { |source_file| source_file.filename =~ /#{filter}/ }
  .group_by(&:gem_name)
```

lib/coverband/utils/result.rb
```diff
@@ -100,6 +100,38 @@ def coverage
         Hash[keys.zip(original_result.values_at(*keys))]
       end
 
+      #
+      # Applies the configured groups to the given array of SimpleCov::SourceFile items
+      #
+      def filter_to_groups(files)
```
kbaum: I thought i had mentioned this but now i can't find my comment so going to add again. I think this method would be helped by introducing a class called FileGroups. Might be a good way to break up the logic.

lib/coverband/reporters/base.rb
danmayer: nice updated

lib/coverband/utils/result.rb
danmayer: agreed as I work on tests I will break this out it is far to complex right now

lib/coverband/utils/result.rb
danmayer: nice

lib/coverband/utils/result.rb
danmayer: yeah this is clean

lib/coverband/utils/result.rb
```diff
@@ -100,6 +100,38 @@ def coverage
...
+        grouped_gems = {}
+        gem_lists = []
+        Coverband.configuration.groups.each do |name, filter|
+          if name == 'Gems'
+            gems_files = files.select { |source_file| source_file.filename =~ /#{filter}/ }
+            gems_files.each do |source_file|
+              grouped_gems[source_file.gem_name] = [] unless grouped_gems[source_file.gem_name]
+              grouped_gems[source_file.gem_name] << source_file
+            end
+            grouped_gems.each_pair do |gem, files|
```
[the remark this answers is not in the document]
danmayer: mostly good had to add flatten

```
    gem_lists = grouped_gems.values.map { |gem_files| Coverband::Utils::FileList.new(gem_files) }
            grouped_files.concat(gem_lists.flatten)
```

lib/coverband/configuration.rb
```diff
@@ -76,6 +81,54 @@ def store=(store)
...
+    end
+
+    def gem_paths
+      # notes ignore any paths that aren't on this system, resolves
+      # bug related to multiple ruby version managers / bad dot files
+      Gem::PathSupport.new(ENV).path.select { |path| File.exist?(path) }
+    end
+
+    SKIPPED_SETTINGS = %w[@s3_secret_access_key @store]
+    def to_h
```
kbaum: Could use each_with_object here?

```ruby
def to_h
  instance_variables.each_with_object({'gem_paths': gem_paths}) do |var, hash|
    hash[var.to_s.delete('@')] = instance_variable_get(var) unless SKIPPED_SETTINGS.include?(var.to_s)
  end
end
```

lib/coverband/utils/file_groups.rb
```diff
@@ -0,0 +1,45 @@
...
+      def grouped_results
+        @grouped
+      end
+
+      private
+
+      def filter_to_groups(files)
+        grouped_files = []
+        grouped_gems = {}
+        gem_lists = []
```
kbaum: Don't think we need to declare gem_lists here?

lib/coverband/utils/file_groups.rb
```diff
@@ -0,0 +1,45 @@
...
+
+#
+# Applies the configured groups to the given array of Coverband::SourceFile items
+#
+module Coverband
+  module Utils
+    class FileGroups
+      def initialize(files)
+        @grouped = {}
+        filter_to_groups(files)
```
kbaum: Thinking that if we set files as state in constructor, then might be able to extract more self documenting methods within this class and reduce the size of the filter_to_groups method. Something like this:

```ruby
def initialize(files)
  @grouped = {}
  @files = files
  filter_to_groups
end
```
Not sure..

lib/coverband/utils/file_groups.rb
```diff
@@ -17,29 +18,36 @@ def grouped_results
...
           else
-            @grouped[name] = Coverband::Utils::FileList.new(files.select do |source_file|
-              source_file.filename =~ /#{filter}/
-            end)
+            app_files(name, filter)
             grouped_files += @grouped[name]
           end
         end
-        if !Coverband.configuration.groups.empty? && !(other_files = files.reject do |source_file|
+        if !Coverband.configuration.groups.empty? && !(other_files = @files.reject do |source_file|
```
kbaum: This method is looking so much simpler! I still feel a little overwhelmed when looking at these last two lines. Perhaps there's a way we can refactor so it's more self documenting? Feel free to disregard as this is definitely ready to :ship.

danmayer:
Thoughts with the 4.0.1 and 4.0.2 should we just merge this in and make it all 4.0.1? I am still waiting to hear back from the folks testing resque, but feels like this is good enough for first release...

Let me know thoughts @kbaum

kbaum:
This looks great! I think we should merge all in and release 4.1.0 (I had released 4.1.0.alpha last week).

danmayer:
agreed @kbaum I am thinking we merge this and the fix for changing directory file merge support (https://github.com/danmayer/coverband/pull/191) and make a 4.1.0.beta release and long as we get not further issues we can push 4.1.0 live this week

PR #184, opened 2019-01-30
Update base key for redis to 4_0

gingerlime:
* Just noticed this when playing around with coverband and inspecting
  redis keys that mismatched my coverband version
* not sure if this is intentional, or what impacts it has on upgrading coverband
  and "losing" the data (although I suspect the data is typically quite short-lived?)

### DIFF
--- a/lib/coverband/adapters/redis_store.rb
+++ b/lib/coverband/adapters/redis_store.rb
@@ -6,7 +6,7 @@ module Coverband
     # RedisStore store a merged coverage file to redis
     ###
     class RedisStore < Base
-      BASE_KEY = 'coverband3_1'
+      BASE_KEY = 'coverband4_0'
 
       def initialize(redis, opts = {})
         super()


kbaum:
So I think this is intentional as when we update that key, it wipes out people's coverage data. We would only do that if we are making a breaking change that makes their existing coverage data incompatible with the new release.

With the latest coverband, each file is reset whenever that specific file changes:

https://github.com/danmayer/coverband#clear-coverage

So people should be able to keep their coverage data forever if that's what they want.

Let me know if that makes sense.

gingerlime:
Makes sense. I guess the version info is a bit unfortunate, but it doesn't really matter in practice. When I saw it, I was a bit unsure that my version was indeed 4.0.1, so it just looks a bit odd :)

kbaum:
It's a very good point that it's confusing. Perhaps we should think about having a separate versioning scheme for the redis data model. We could use a checksum that's we commit into source code in the version.rb file. Then we only change that checksum when we know we've made a backwards incompatible data format change.

danmayer:
yeah if we change the constant name or add a comment we could clarify... While the migrations are interesting I feel like they would be a lot of work to maintain and with little value as most folks are OK with clearing / resetting coverage from time to time. 

I think in terms of simplicity I would just go with changing constant name / version number to to not match coverband release version at all and document it is the version the internal redis format / structure used by coverband.

gingerlime:
If people are ok with clearing / resetting coverage, then this breaking change is quite minor. And they can manually "migrate" if they want to keep their data... 

Going forward, I think it's useful to keep some kind of versioning of data... especially if at some point the data representation changes, so it can be useful to know which version of data you're looking at (but not sure how likely this is going to be).

I find it confusing to have used a versioned base key, but then not keeping up the versioning. But anyway, it's no biggie. Just my 2 cents.

danmayer:
OK I made these changes take a look @gingerlime https://github.com/danmayer/coverband/commit/a09e2ad4f972383137780a8a59773902df50b512

PR #192, opened 2019-02-11
Change background reporting frequency

kbaum:


kbaum:
Did some more experimentation on this on the morning commute. Noticing that `test/unit/rails_full_stack_test.rb` was failing every 10th time. By including as a reload_file, was able to execute 50+ times in a row with no failure.  My theory is that in some cases, the dummy_controller was being required before the `Coverage.start` which causes class load coverage to be missed.

Also added safe_reload_files to reset method in Configuration.

danmayer:
Nice well done @kbaum thanks for digging in

kbaum:
Looks like there are still some failures now. Will dig in some more later on.

kbaum:
Closing as this is my new theory:
https://github.com/danmayer/coverband/pull/194

PR #194, opened 2019-02-12
Manually report coverage in test

kbaum:
Modified these two tests to not use the background reporting thread and it seems to fix the sporadic failures. Working theory which i haven't had time to verify. The current minitest thread has a background thread running. Then this test spins up a puma server which might fork and start another background thread. These threads compete with each other and cause occasional overwrites of data.

We could merge this in and I believe it will solve sporadic failures for these tests. The other alternative is to dig in some more here and:

1. See if my theory is actually correct
2. Fix by stopping the thread in the minitest process?

This pull request takes a little bit of "end to end" away in that it does not test the background thread so i am on the fence.


danmayer:
Yeah, let's merge this testing background threads is obviously a bit tricky and we can work on some more isolated ways to do that.

Issue #193, opened 2019-02-12
data size can balloon under certain conditions


gingerlime:
Relates to the discussion on #188.

If the folder structure of the app changes when code is deployed (e.g. incremental release id sub-folders), the data size can grow unbounded, with every deploy.

This is due to the fact that keys are based on the full path.

There are a few options for resolving this (still in discussion):

* when [merging](https://github.com/danmayer/coverband/blob/2f0785eac742c0d28d2e55761f78eb573f309712/lib/coverband/adapters/base.rb#L64) coverage data, merge files that have the same md5 signature (this will keep old code versions though, but not all files)
* canonicalize the full path when storing file keys (similar to what's happening with `root_paths`, would incur a fixed "penalty" for each key, but depending on the complexity of the `root_paths` regex complexity)
* create a rake task to clean-up duplicates (based on which criteria?)
* something else?

danmayer:
yeah, I like the idea of using relative paths while reporting opposed to merging... Both solutions would generally be in the background thread and shouldn't case to much of a performance issue.

@kbaum I think you also preferred relative paths vs merging.

One thing to note, is this likely doesn't work for gem and stdlib Ruby files as their paths aren't always relative to the app.

* gems we could do relative to the gem_home, or we could do nothing and assume it doesn't change per deploy
* std-lib while Coverage tracks at the moment Coverband doesn't support or care about reporting std-lib so they should just be ignored and never make it to the store.

danmayer:
thoughts @gingerlime / @kbaum I can start working on an implementation.

kbaum:
I think relative paths within our store (redis) is a solid solution. I don't think performance should be an issue either. Could always optimize with an in memory cache of absolute path to relative if it became an issue.

gingerlime:
just to clarify, reporting means when coverage data gets pulled from coverage and stored? (as opposed to the web interface report). Sorry, I'm not entirely familiar with your terminology :)

Yes, intuitively it makes the most sense to me. 

The only downside I can think of is if you change your `root_paths` regex at some point, you also need to clear your data, because it won't be able to "backport" things if there was already a regex applied. It's a very odd edge case though, so not something to worry about in my opinion (and easily "fixed" by clearing data). 

👍 from me!

danmayer:
[...]

PR #195, opened 2019-02-14
move store to use relative paths, this handles when paths are differe…

danmayer:
…nt on different machines or the directories change during deploys.

This addresses #193 

* at the moment it is set up to break old data
  * we could just release it breaking previous coverage
  * we could have a rake task to migrate previous data (easiest and most performant)
          * this seems like the best model going forward, we could always support migrating from only one previous data store version
  * we could auto-detect and update, the problem is it would add permanent overhead to the merge process. This also muddies up the tight loop runtime code.

I could still clean this up a bit more and refactor some of the tests to more directly hit the file_path_helpers opposed to only through their usage methods


lib/coverband/adapters/base.rb
```diff
@@ -57,8 +60,9 @@ def expand_report(report)
             'file_hash' => file_hash(key),
             'data' => line_data
           }
-          report[key] = extended_data
+          expanded[full_path_to_relative(key)] = extended_data
```
kbaum: When I first looked at the concepts here I got a little confused. We are using the word expanded to expand the data we are storing but converting the files from full path to relative path. Possible sign that this method is doing more than 1 thing or that my brain can't handle very much 😏

lib/coverband/adapters/redis_store.rb
```diff
@@ -11,7 +11,7 @@ class RedisStore < Base
       # used to store data to redis. It is changed only when breaking changes to our
       # redis format are required.
       ###
-      REDIS_STORAGE_FORMAT_VERSION = 'coverband3_1'
+      REDIS_STORAGE_FORMAT_VERSION = 'coverband_3_2'
```
kbaum: I know we talked about this before but this is going to be confusing for us to keep track of in our brains. We do always have git but we might want to think about just making this match the coverband version number whenever we bump this format version.

lib/coverband/reporters/base.rb
```diff
@@ -20,25 +21,14 @@ def report(store, _options = {})
...
-
         def fix_file_names(report_hash, roots)
           if Coverband.configuration.verbose
             Coverband.configuration.logger.info "fixing root: #{roots.join(', ')}"
           end
 
           # normalize names across servers
           report_hash.each_with_object({}) do |(key, vals), fixed_report|
-            filename = filename_from_key(key, roots)
+            filename = relative_path_to_full(key, roots)
```
kbaum: I'm surprised we have to convert back from relative to absolute? I pictured this only going one way.

lib/coverband/utils/file_path_helper.rb
```diff
@@ -0,0 +1,68 @@
...
+      #          (if using \d for example)
+      #        or use single qoutes
+      #        example: '/box/apps/app_name/releases/\d+/'
+      #        example: '/var/local/company/company.d/[0-9]*/'
+      ###
+      def relative_path_to_full(relative_path, roots)
+        relative_filename = relative_path
+        local_filename = relative_filename
+        roots.each do |root|
+          relative_filename = relative_filename.gsub(/^#{root}/, './')
```
kbaum: If this file is relative, why do we need to chop off the root?

lib/coverband/utils/file_path_helper.rb
```diff
@@ -0,0 +1,68 @@
...
+      # NOTEs on configuration.root_paths usage
+      # strings: matching is pretty simple for full string paths
+      # regex: to get regex to work for changing deploy directories
+      #        the regex must be double escaped in double quotes
+      #          (if using \d for example)
+      #        or use single qoutes
+      #        example: '/box/apps/app_name/releases/\d+/'
+      #        example: '/var/local/company/company.d/[0-9]*/'
+      ###
+      def relative_path_to_full(relative_path, roots)
```
kbaum: Would something like this work here?
```
irb(main):001:0> File.expand_path('Gemfile')
=> "/Users/karlbaum/workspace/coverband/Gemfile"
```

lib/coverband/utils/file_path_helper.rb
```diff
@@ -0,0 +1,68 @@
...
+        # a fully expanded path of config.root
+        # filename = filename.gsub('./', roots.last)
+        # above only works for app files
+        # we need to rethink some of this logic
+        # gems aren't at project root and can have multiple locations
+        local_root = roots.find { |root| File.exist?(relative_filename.gsub('./', root)) }
+        local_root ? relative_filename.gsub('./', local_root) : local_filename
+      end
+
+      def root_paths
```
kbaum: I feel like this belongs on Coverband itself or the Coverband::Configuration and should be initialized once?

lib/coverband/utils/file_path_helper.rb
```diff
@@ -0,0 +1,68 @@
...
+        roots = Coverband.configuration.root_paths
+        roots += Coverband.configuration.gem_paths if Coverband.configuration.track_gems
+        roots << "#{current_root}/"
+        roots
+      end
+
+      def current_root
```
kbaum: Similar comment here: https://github.com/danmayer/coverband/pull/195/files#r256836620

lib/coverband/adapters/base.rb
danmayer: yeah, this in our tight reporting loop I am trying to stay with as few iterations as possible through the collection. As this method has large impact on the perf. I think renaming the method could help clarify but hard to avoid doing more than one thing

lib/coverband/adapters/redis_store.rb
danmayer: I think it is worth keeping it independent so we can change the version differently than our gem version. We renamed the constant to improve that confusion. I think we could try to change the numbering significantly to help as well. If we are going to introduce the concept of migrations and such I really don't want it tied at all to Coverband version as this is far more stable.

I am starting to think about it as a schema_version, we could use a timestamp to match Rails schema but the timestamp wouldn't be user generate but updated when we are forcing a version change?

lib/coverband/reporters/base.rb
danmayer: all the web reporting / HTML expects a full path. Part of this is legacy as SimpleCov always expected full path. So further refactorings are likely possible but it would touch most if not all the SimpleCov files so it is a fairly major change.

danmayer: Also not true for Gems which aren't relative paths

lib/coverband/utils/file_path_helper.rb
danmayer: Not true for Gem data, gems have gem_home which is one of the roots

lib/coverband/utils/file_path_helper.rb
danmayer: needs to support Gems

lib/coverband/utils/file_path_helper.rb
danmayer: yeah I thought about it as well I kept it here since it was all paths, but happy to move it there

lib/coverband/utils/file_path_helper.rb
danmayer: sorry this doesn't link to a comment, not sure what you mean

lib/coverband/adapters/redis_store.rb
kbaum: I like the idea of modeling after rails migrations and makes sense that we would manually update it as this doesn't happen too often.

lib/coverband/utils/file_path_helper.rb
kbaum: Makes sense. I can see that. I usually avoid modules for code reuse. Prefer to have methods that belong to objects for some reason. I find that going overboard with modules for organizing code can make things confusing and each one starts to feel like a junk drawer.

lib/coverband/utils/file_path_helper.rb
kbaum: Same comment about this method belonging on Coverband or Coverband::Configuration

lib/coverband/utils/file_path_helper.rb
```diff
@@ -0,0 +1,68 @@
...
+          relative_filename = relative_filename.gsub(/^#{root}/, './')
+        end
+        # the filename for our reports is expected to be a full path.
+        # roots.last should be roots << current_root}/
+        # a fully expanded path of config.root
+        # filename = filename.gsub('./', roots.last)
+        # above only works for app files
+        # we need to rethink some of this logic
+        # gems aren't at project root and can have multiple locations
+        local_root = roots.find { |root| File.exist?(relative_filename.gsub('./', root)) }
```
gingerlime: If roots contain a regex, then wouldn't `File.exist?` fail? Also, this probably incurs a potentially big performance hit on I/O ...

lib/coverband/reporters/base.rb
gingerlime: I made a comment below, because I'm not sure that the reverse is even technically possible in all cases, especailly with regex... 

But besides that, I'm not sure if I understand why we care for the path to be "real" real, rather than normalized / canonicalized. Does it mean SimpleCov won't be able to open our file unless we give it the full path? (sorry, no clue about SimpleCov or where it's used etc)

lib/coverband/adapters/redis_store.rb
gingerlime: I think it exactly serves its purpose here. If the data structure changes in a non-compatible way, this is the way to manage it without creating incompatibility.

Of course, if people want to keep their data on upgrade, they can write their own migrations, or the project can decide it's worthwhile providing those migrations for them. I think that in the majority of cases, the data is by-and-large disposable. (of course, people want to keep it, but telling them "this upgrade will wipe your coverage data, unless you do X" seems reasonable to me).

lib/coverband/utils/file_path_helper.rb
gingerlime: I see your point about the junk drawer, but it doesn't feel this way to me. The helper is clearly named, the methods here are focused on paths and not something else. This feels like a modular codebase to me personally and quite a way off from being a junk drawer (although some junk drawers have their purpose too, but let's not go too far there...)

lib/coverband/utils/file_path_helper.rb
gingerlime: Sorry, I didn't realize that it was already doing this `File.exist?` before ... So feel free to ignore. I'm still a little confused on how it would reverse a regex and find the full path though (took a look at the tests, but not sure I completely grok it either). Sorry, it's mostly my lack of familiarity, and admittedly I spent only a tiny bit of time on this.

gingerlime:
I'd be happy to test it on my staging environment if/when it's ready. Would be great if you guys can push a beta release, but otherwise, I can also probably pull the gem from github if necessary

lib/coverband/utils/file_path_helper.rb
kbaum: For sure this is not a "junk drawer" and I probably shouldn't have used those words. I think in general this is a module with a bunch of helper methods that are similar in that they have to do file system stuff. This is a fine way to organize these methods. In general I prefer to have methods defined on objects if possible. If they are pure helper functions with no state and don't belong on an object, then I prefer making methods on the module object like:

```ruby
def self.relative_path_to_full(relative_path, roots)
```
This I find to be more self documenting regarding how this method is to be used..

```
FilePathHelper.relative_path_to_full('./foo/bar.rb', '/app/')
```

Hope that makes sense and in this case, I think it's fine to leave as is.

lib/coverband/adapters/redis_store.rb
kbaum: What I like about the migration rake task with the `REDIS_STORAGE_FORMAT_VERSION` change is that it doesn't really wipe the data. It simply starts storing the data in a different spot. A likely scenario is that people will upgrade coverband and go to view the report and see their data is gone. If they care, they will either go to our changelog or README and see instructions there or open an issue on github and we can instruct them. In the end, they can recover their data if they want to.

gingerlime: Leaving stale data behind isn't great either, though. But I agree with you...

kbaum: True but I think coverband doesn't take up too much space and the default ttl should clean things up eventually.

gingerlime:
Hey @danmayer. I'm testing this on staging, and it looks good!

I did observe one weird thing however. when I go to the settings page, here's the list of root_paths it shows (sanitized)

```
root_paths
    ["/var/local/company/company.d/[0-9]*/", "/var/local/company/company.d/111/", ..., "/var/local/company/company.d/111/"]
```

It's literally hundreds of entries. I'm guessing one for each source file (?) but not entirely sure.

danmayer:
Hmm ok interesting I will look.

I am traveling so a bit slow to close this out but should get to it soon
ish.

danmayer:
The duplicates in root_paths is definitely a bug right now. I will address soon hopefully will have time tomorrow.

lib/coverband/adapters/redis_store.rb
danmayer: Also, rake migrations can wipe old data after the update.

lib/coverband/utils/file_path_helper.rb
danmayer: yeah this could be simplified further all the regex ones fail until it finds the valid ones which will either be the current_root or the gem_paths which were added to root paths. So this wastes some time during reporting but is safe.

lib/coverband/utils/file_path_helper.rb
danmayer: OK moved two of the helpers to config and kept the others here for now.

lib/coverband/utils/file_path_helper.rb
danmayer: OK moved two of the helpers to config and kept the others here for now.

danmayer:
fixed the root paths issue

PR #196, opened 2019-02-19
Stop parent resque background reporting thread

kbaum:
No need for the background thread to be running in the parent resque process. This stops the background thread before the first fork.


danmayer:
cool makes sense, good once tests pass

kbaum:
Cool.. i think this commit fixed the sporadic failure:

https://github.com/danmayer/coverband/pull/196/commits/c920c0b7bb2e9b5f10c863cc777d08e4c7dc3d6f

Stops background thread from starting in the child process.

PR #202, opened 2019-03-03
Update 'Collecting Gem / Library Usage' section

kbaum:
Took a stab at updating the readme here. Things got a bit complicated when I started thinking about the order in which gems need to be required. Realizing that gem require ordering is quite confusing for coverband. For example, I think the resque integration would not work at all if coverband was required before resque.

I also didn't mention anything about using reload files to track gems required before coverband such as rails. Thinking I might want to add that in but just wanted to open this up to get feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,16 +251,28 @@ By adding any files above you will get reporting on those files as part of your 
 
 ### Collecting Gem / Library Usage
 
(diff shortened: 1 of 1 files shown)


danmayer:
yeah you are right about Resque, also not sure about Railties working if required before Rails... I guess this also relates to the recent issue where someone didn't have their ENV gem loaded before Coverband so it couldn't find their REDIS_URL...

Anyways, this looks great for a first pass, thanks. +1 to merging it.

kbaum:
Ok thanks. Will merge this for now and we can iterate on it.

Issue #186, opened 2019-02-01
Eager loading of files skews the coverage percentages


CharlesMcMillan:
In our coverband results, we are often seeing files that are mostly unused, but showing percentages above 60% or so...the reason being that the method definition `def method_name` and `end` are counted as being covered...which makes sense because the file is loaded.

In production, we use `eager_load = true` on our rails server, and I _think_ that might be related to what we're seeing here.

Is there some way that coverband could filter out the lines that are apart of the method definition (like `def method_name`) so that the percentages aren't skewed to be so high?

danmayer:
yeah I have been thinking about this a lot lately @CharlesMcMillan ... it is good to know they are run... but it is very different what is executed at load time vs run-time... Basically, the standard-lib of Coverage doesn't know the difference, but I think Coverband could figure this out by diffing eager load lines... to what it gets past initial file load during runtime. This kind of difference also needs to show when a file last had any usage in Coverband right now the last_file_activity metric shows last time file loaded, which isn't helpful.

Thanks for the report, I am thinking about this and have some ideas, but I think it will take a bit longer to work out a good solution. Not likely to make it in the 4.0.2 release but this might be a good primary feature to drive our 4.0.3 release.

kbaum:
@danmayer That sounds like a great way to handle it. Another thought i had here is that what we really want to know is percent method body execution. I wonder if there is a way to calculate that?

danmayer:
hmm could you explain more... would this be kind of like a new breakdown of the file... right now we have relevant lines which is non-white-space, non comment, etc lines... is 'method body' kind of like even more interesting relevant lines?

Are method body lines somehow different than `on load` / when required lines... or just better naming?

I think the idea is good something like below...

existing
`| % coverage | lines | RELEVANT LINES | Lines Covered | Lines Missed |`

new
`| % coverage | % BODY COVERAGE | lines | RELEVANT LINES | METHOD BODY LINES | Lines Covered | Lines Missed |`

kbaum:
My last idea was not well thought out. I was just thinking that most often method bodies are missed and only class definitions such as constants, macros, and method signatures are hit on app load. This is not 100% true though and aside from that it would probably be difficult for coverband to know which lines of code are inside a method body.

[...]

PR #205, opened 2019-03-11
Reset coverage before each test

kbaum:
Depending on which order the tests are run in the current coverage could be different. This PR introduces a call to Coverband#result which should clear what is in ::Coverage::peek_result

### DIFF
--- a/test/coverband/collectors/coverage_test.rb
+++ b/test/coverband/collectors/coverage_test.rb
@@ -1,53 +1,72 @@
 # frozen_string_literal: true
 
 require File.expand_path('../../test_helper', File.dirname(__FILE__))
-require File.expand_path('../../dog', File.dirname(__FILE__))
 
-if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.3.0')
-  class CollectorsCoverageTest < Minitest::Test
-    attr_accessor :coverband
+class CollectorsCoverageTest < Minitest::Test
+  attr_accessor :coverband
 
-    def setup
-      super
-      Coverband.configure do |config|
-        config.store = Coverband::Adapters::RedisStore.new(Redis.new)
-      end
-      @coverband = Coverband::Collectors::Coverage.instance.reset_instance
+  def setup
+    super
+    Coverband.configure do |config|
+      config.store = Coverband::Adapters::RedisStore.new(Redis.new)
     end
+    @coverband = Coverband::Collectors::Coverage.instance.reset_instance
+  end
 
-    def teardown
-      Thread.current[:coverband_instance] = nil
-      Coverband.configure do |config|
-      end
-      @coverband = Coverband::Collectors::Coverage.instance.reset_instance
+  def teardown
+    Thread.current[:coverband_instance] = nil
+    Coverband.configure do |config|
     end
+    @coverband = Coverband::Collectors::Coverage.instance.reset_instance
+  end
 
-    test 'gets coverage instance' do
-      assert_equal Coverband::Collectors::Coverage, coverband.class
-    end
+  test 'Empty coverage' do
+    coverband.report_coverage(true)
+    assert_equal({}, Coverband.configuration.store.coverage)
+  end
 
-    test 'defaults to a redis store' do
-      assert_equal Coverband::Adapters::RedisStore, coverband.instance_variable_get('@store').class
-    end
+  test 'Dog class coverage' do
+    load File.expand_path('../../dog.rb', File.dirname(__FILE__))
+    coverband.report_coverage(true)
+    coverage = Coverband.configuration.store.coverage
+    assert_equal(["./test/dog.rb"], coverage.keys)
+    assert_equal(coverage["./test/dog.rb"]["data"], [nil, nil, 1, 1, 0, nil, nil])
+  end
 
-    test 'report_coverage raises errors in tests' do
-      @coverband.reset_instance
-      @coverband.expects(:ready_to_report?).raises('Oh no')
-      assert_raises RuntimeError do
-        @coverband.report_coverage
-      end
-    end
+  test 'Dog method and class coverage' do
+    load File.expand_path('../../dog.rb', File.dirname(__FILE__))
+    Dog.new.bark
+    coverband.report_coverage(true)
+    coverage = Coverband.configuration.store.coverage
+    assert_equal(["./test/dog.rb"], coverage.keys)
+    assert_equal(coverage["./test/dog.rb"]["data"], [nil, nil, 1, 1, 1, nil, nil])
+  end
 
-    test 'report_coverage does not raise errors in non-test mode' do
-      Coverband.configuration.stubs(:test_env).returns(false)
-      @coverband.expects(:ready_to_report?).raises('Oh no')
-      @coverband.reset_instance
+  test 'gets coverage instance' do
+    assert_equal Coverband::Collectors::Coverage, coverband.class
+  end
+
+  test 'defaults to a redis store' do
+    assert_equal Coverband::Adapters::RedisStore, coverband.instance_variable_get('@store').class
+  end
+
+  test 'report_coverage raises errors in tests' do
+    @coverband.reset_instance
+    @coverband.expects(:ready_to_report?).raises('Oh no')
+    assert_raises RuntimeError do
       @coverband.report_coverage
     end
+  end
 
-    test 'default tmp ignores' do
-      heroku_build_file = '/tmp/build_81feca8c72366e4edf020dc6f1937485/config/initializers/assets.rb'
-      assert_equal false, @coverband.send(:track_file?, heroku_build_file)
-    end
+  test 'report_coverage does not raise errors in non-test mode' do
+    Coverband.configuration.stubs(:test_env).returns(false)
+    @coverband.expects(:ready_to_report?).raises('Oh no')
+    @coverband.reset_instance
+    @coverband.report_coverage
+  end
+
+  test 'default tmp ignores' do
+    heroku_build_file = '/tmp/build_81feca8c72366e4edf020dc6f1937485/config/initializers/assets.rb'
+    assert_equal false, @coverband.send(:track_file?, heroku_build_file)
   end
 end
--- a/test/integration/rails_gems_full_stack_test.rb
+++ b/test/integration/rails_gems_full_stack_test.rb
@@ -11,10 +11,11 @@ class RailsGemsFullStackTest < Minitest::Test
     # The normal relative directory lookup of coverband won't work for our dummy rails project
     Coverband.configure("./test/rails#{Rails::VERSION::MAJOR}_dummy/config/coverband.rb")
     Coverband.configuration.track_gems = true
-    Coverband.configuration.gem_details = false
+    Coverband.configuration.gem_details = true
     Coverband.configuration.background_reporting_enabled = false
     Coverband.start
     require 'rainbow'
+    load('rainbow/wrapper.rb')
     Rainbow('this text is red').red
   end
 
--- a/test/integration/full_stack_test.rb
+++ b/test/integration/full_stack_test.rb
@@ -44,6 +44,7 @@ class FullStackTest < Minitest::Test
     Coverband.configuration.track_gems = true
     Coverband::Collectors::Coverage.instance.reset_instance
     require 'rainbow'
+    load 'rainbow/null_presenter.rb'
     Rainbow('this text is red').red
     request = Rack::MockRequest.env_for('/anything.json')
     middleware = Coverband::Middleware.new(fake_app_with_lines)
--- a/test/integration/rails_full_stack_test.rb
+++ b/test/integration/rails_full_stack_test.rb
@@ -12,6 +12,7 @@ class RailsFullStackTest < Minitest::Test
     Coverband.configure("./test/rails#{Rails::VERSION::MAJOR}_dummy/config/coverband.rb")
     Coverband.configuration.background_reporting_enabled = false
     Coverband.start
+    load "./test/rails#{Rails::VERSION::MAJOR}_dummy/app/controllers/dummy_controller.rb"
   end
 
   def teardown
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -23,6 +23,11 @@ Coveralls.wear!
 module Coverband
   module Test
     def self.reset
+      if defined?(::Coverage.running?)
+        ::Coverage.result if ::Coverage.running?
+      else
+        ::Coverage.result rescue nil
+      end
       Coverband.configuration.store.clear!
       Coverband.configuration.reset
       Coverband::Collectors::Coverage.instance.reset_instance
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -84,18 +84,6 @@ def basic_coverage
   { 'app_path/dog.rb' => example_line }
 end
 
-def fake_redis
-  @redis ||= begin
-    redis = OpenStruct.new
-    redis
-  end
-end
-
-def fake_coverage_report
-  file_name = '/Users/danmayer/projects/hearno/script/tester.rb'
-  { file_name => [1, nil, 1, 1, nil, nil, nil] }
-end
-
 def source_fixture(filename)
   File.expand_path(File.join(File.dirname(__FILE__), 'fixtures', filename))
 end


test/integration/rails_gems_full_stack_test.rb
```diff
@@ -11,10 +11,11 @@ def setup
     # The normal relative directory lookup of coverband won't work for our dummy rails project
     Coverband.configure("./test/rails#{Rails::VERSION::MAJOR}_dummy/config/coverband.rb")
     Coverband.configuration.track_gems = true
-    Coverband.configuration.gem_details = false
+    Coverband.configuration.gem_details = true
```
kbaum: I'm not sure how this test was working with `gem_details = false` as it seems to be looking for coverage on a specific file within the gem.

kbaum:
Actually... i think this may introduce an unintended side effect:

image

kbaum:
Looks like coveralls uses simplecov which unsurprisingly has a dependency on ruby ::Coverage:

https://github.com/colszowka/simplecov/blob/master/lib/simplecov.rb#L46

Clearing coverage after each test call might be a lost cause.

kbaum:
Opening this up again as I think it's worth talking about not having coveralls reporting in order to clear Coverage#peek_result before a test is run. My thought is that this kind of like having a clean database. Leaving coverage data around between tests allows for test ordering dependency issues.

test/coverband/collectors/coverage_test.rb
```diff
@@ -1,53 +1,72 @@
 # frozen_string_literal: true
 
 require File.expand_path('../../test_helper', File.dirname(__FILE__))
-require File.expand_path('../../dog', File.dirname(__FILE__))
-
-if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.3.0')
```
kbaum: Removed this statement now that we only support >= 2.3.0.

test/coverband/collectors/coverage_test.rb
```diff
@@ -1,53 +1,72 @@
...
+  end
 
-    test 'report_coverage raises errors in tests' do
-      @coverband.reset_instance
-      @coverband.expects(:ready_to_report?).raises('Oh no')
-      assert_raises RuntimeError do
-        @coverband.report_coverage
-      end
-    end
+  test 'Empty coverage' do
```
kbaum: Added a few basic tests here.

test/integration/rails_gems_full_stack_test.rb
danmayer: hmmm that is weird.

kbaum: My guess is that it was picking up coverage from a different test?

danmayer: hmm yeah cause without gem details it should render the gem name but not the individual files... makes me think I am likely still adding too much unneeded HTML to the report when gem defails is false.

danmayer:
I guess what kind of errors do we think we will have... I get your point about not clearing Coverage being like a database, but Ruby stdlib Coverage doesn't allow a real stop and start, so you really can't clear the Coverage lib, you can only reset what the app was comparing it to... 

If we really want a CLEAR coverage, I would go back to the idea of running every spec in an isolated process. If Coverage was started and run per process that should also isolate it. That could be overkill...

In that sense though, I feel like ensuring we have good coverage of Coverband itself is a more real issue than test dependency issues. If we really wanted to avoid that we could actually turn off running tests in a random order. Our test suite is small enough that we are unlikely to be splitting it to run in parallel, and I don't think it would introduce to big of an issue to document tests are run in an expected order... The test run order issue right now just resulted in slightly different Coverage results... The new one we get zero, but I don't think it highlights any real bug in the library only possibly tests that aren't as good as we hoped.

Thoughts?

kbaum:
> Ruby stdlib Coverage doesn't allow a real stop and start, so you really can't clear the Coverage lib, you can only reset what the app was comparing it to...

Put together a little script to test it out:

```ruby
#!/usr/bin/env ruby

require 'coverage'
Coverage.start
load './dog.rb'
Dog.new.bark
puts Coverage.peek_result
Coverage.result
Coverage.start
puts Coverage.peek_result
```
And I get:
```
{"/Users/karlbaum/workspace/coverband/test/dog.rb"=>[nil, nil, 1, 1, 1, nil, nil]}
{}
```
This seems like it's clearing the ruby coverage data to me but let me know if I am missing something. One note, I noticed that when I run in ruby 2.3, only the data is cleared and not the file name. In ruby 2.3, when I run the script, I get:

```
{"/Users/karlbaum/workspace/coverband/test/dog.rb"=>[nil, nil, 1, 1, 1, nil, nil]}
{"/Users/karlbaum/workspace/coverband/test/dog.rb"=>[]}
```
ruby 2.4+ seems to clear everything as in the ruby 2.5 run.

> If we really want a CLEAR coverage, I would go back to the idea of running every spec in an isolated process. If Coverage was started and run per process that should also isolate it. That could be overkill...

This is a very interesting idea. I could get behind that if resetting the Coverage data was not an option.

> and I don't think it would introduce to big of an issue to document tests are run in an expected order... 

I hadn't considered actually running the tests in a certain order and it is an interesting idea. I think introducing order dependencies might be confusing for engineers trying to contribute to the project. It's just not something that most developers would be used to. 

> The test run order issue right now just resulted in slightly different Coverage results... The new one we get zero, but I don't think it highlights any real bug in the library only possibly tests that aren't as good as we hoped.

I definitely did not do a great job in explaining the issue with the coveralls test coverage. This PR introduces an incompatibility with coveralls test coverage because I introduced the call to ```Coverage.result``` before each test here:

https://github.com/danmayer/coverband/pull/205/files#diff-f14ca9959ff06f3ca2ee6b9564866954R27

Coveralls which depends on simplecov which depends on ruby coverage no longer has any coverage data since each test run it gets cleared in the setup. Hope that makes sense.

danmayer:
If you dig into a bunch of the ruby bugtracker related to Coverage, it clears the memory (hence the data or both data and name going away). Then it adds files to various ignores, but Ruby has already "compiled" those files differently with all the Coverage hooks. If you stop then hit a file again it won't show new coverage... If you call load again, it would then work again.

So yes you can stop and start... but it doesn't quite work exactly how one would expect as most Ruby projects require everything once at start and don't use load.

Basically, we have used the 'load' trick a number of places in tests to get Coverage to track a file while a normal `require 'somefile.rb'`...

So we could under the hood call `Coverage.result` in our tests and use load to track thing as needed. Still losing all coverage data on our actual tests seems worse than the original problem of possible ordering issues.

I guess what is the real problem you are seeing or trying to solve and perhaps we can go about it another way?

>  I introduced the call to Coverage.result before each test here...

This was introduced to solve, slight variability in Coverage?

* if we don't want to keep tests in order, what causes that variability? (a require, something clearning, etc?)
* could we have each test that checks coverage use a uniq tmp file per test? isolating that way?
* test coverage on files that aren't 'counted' as part of our apps coverage?

danmayer:
Another idea would be to isolate our tests from Coverage by mocking / stubbing it out as if it was a 3rd party service. Could even have something like VCR to record Coverage...

kbaum:
> Basically, we have used the 'load' trick a number of places in tests to get Coverage to track a file while a normal require 'somefile.rb'...

True, though this does not feel unexpected as calling require is a noop if the file has already been required.

> Still losing all coverage data on our actual tests seems worse than the original problem of possible ordering issues.

Agreed I also like the idea of being able to track our test coverage. Though, it wouldn't be surprising that a project like coverband which has such heavy dependency on `::Coverage` itself would run into an issue like this. 

> I guess what is the real problem you are seeing or trying to solve and perhaps we can go about it another way?

Great question. I actually sat down to work on [this issue](https://github.com/danmayer/coverband/issues/186). I started at it from the test side and I think I was bothered that when I looked into what was stored in `Coverband.configuration.store.coverage` I saw bunch of other files that were unrelated to my test.

For example, I introduced the following example into coverage_test.rb in this PR (Added puts for this example):

```ruby
  test 'Empty coverage' do
    coverband.report_coverage(true)
    puts "coverage length: #{Coverband.configuration.store.coverage.keys.length}, files: #{Coverband.configuration.store.coverage.keys.join("\n")}"
    assert_equal({}, Coverband.configuration.store.coverage)
  end
```
That test is very simple. If we haven't called any code, there should be an empty report. That works if we reset Coverage before each test, but if I remove that reset it fails as ` Coverband.configuration.store.coverage` comes back with data. Run just that test:

```
bundle exec m test/coverband/collectors/coverage_test.rb:23
```
I see this on console

```
coverage length: 24, files: ./lib/coverband.rb
./lib/coverband/at_exit.rb
./lib/coverband/configuration.rb
./lib/coverband/utils/file_path_helper.rb
./lib/coverband/adapters/base.rb
./lib/coverband/adapters/redis_store.rb
./lib/coverband/adapters/file_store.rb
./lib/coverband/utils/s3_report.rb
./lib/coverband/utils/html_formatter.rb
./lib/coverband/utils/result.rb
./lib/coverband/utils/file_list.rb
./lib/coverband/utils/gem_list.rb
./lib/coverband/utils/source_file.rb
./lib/coverband/utils/file_groups.rb
./lib/coverband/utils/lines_classifier.rb
./lib/coverband/collectors/coverage.rb
./lib/coverband/reporters/base.rb
./lib/coverband/reporters/html_report.rb
./lib/coverband/reporters/console_report.rb
./lib/coverband/integrations/background.rb
./lib/coverband/integrations/rack_server_check.rb
./lib/coverband/reporters/web.rb
./lib/coverband/integrations/middleware.rb
./lib/coverband/integrations/resque.rb
```
Kind of expected I guess because we see coverband files. If I run the full test suite:

```
bundle exec rake test
```

I see this:

```
coverage length: 57, files: ./lib/coverband.rb
./lib/coverband/at_exit.rb
./lib/coverband/configuration.rb
./lib/coverband/utils/file_path_helper.rb
./lib/coverband/adapters/base.rb
./lib/coverband/adapters/redis_store.rb
./lib/coverband/adapters/file_store.rb
./lib/coverband/utils/s3_report.rb
./lib/coverband/utils/html_formatter.rb
./lib/coverband/utils/result.rb
./lib/coverband/utils/file_list.rb
./lib/coverband/utils/gem_list.rb
./lib/coverband/utils/source_file.rb
./lib/coverband/utils/file_groups.rb
./lib/coverband/utils/lines_classifier.rb
./lib/coverband/collectors/coverage.rb
./lib/coverband/reporters/base.rb
./lib/coverband/reporters/html_report.rb
./lib/coverband/reporters/console_report.rb
./lib/coverband/integrations/background.rb
./lib/coverband/integrations/rack_server_check.rb
./lib/coverband/reporters/web.rb
./lib/coverband/integrations/middleware.rb
./lib/coverband/integrations/resque.rb
./test/coverband/adapters/file_store_test.rb
./test/coverband/adapters/redis_store_test.rb
./test/coverband/at_exit_test.rb
./test/coverband/collectors/coverage_test.rb
./test/coverband/configuration_test.rb
./test/coverband/coverband_test.rb
./test/coverband/integrations/background_test.rb
./test/coverband/integrations/middleware_test.rb
./test/coverband/integrations/rack_server_check_test.rb
./test/coverband/integrations/resque_worker_test.rb
./test/coverband/reporters/base_test.rb
./test/coverband/reporters/console_test.rb
./test/coverband/reporters/html_test.rb
./test/coverband/reporters/web_test.rb
./test/coverband/utils/file_groups_test.rb
./test/coverband/utils/file_list_test.rb
./test/coverband/utils/gem_list_test.rb
./test/coverband/utils/lines_classifier_test.rb
./test/coverband/utils/result_test.rb
./test/coverband/utils/s3_report_test.rb
./test/coverband/utils/source_file_line_test.rb
./test/coverband/utils/source_file_test.rb
./test/integration/full_stack_test.rb
./test/integration/rails_full_stack_test.rb
./test/rails_test_helper.rb
./test/rails5_dummy/config/environment.rb
./test/rails5_dummy/config/application.rb
./test/rails5_dummy/config/coverband.rb
./test/rails5_dummy/config/routes.rb
./test/integration/rails_gems_full_stack_test.rb
./test/rails5_dummy/app/controllers/dummy_controller.rb
./test/coverband/integrations/test_resque_job.rb
./test/dog.rb
```
Run `bundle exec rake test` one more time and it's slightly different:

```
coverage length: 58, files: ./lib/coverband.rb
./lib/coverband/at_exit.rb
./lib/coverband/configuration.rb
./lib/coverband/utils/file_path_helper.rb
./lib/coverband/adapters/base.rb
./lib/coverband/adapters/redis_store.rb
./lib/coverband/adapters/file_store.rb
./lib/coverband/utils/s3_report.rb
./lib/coverband/utils/html_formatter.rb
./lib/coverband/utils/result.rb
./lib/coverband/utils/file_list.rb
./lib/coverband/utils/gem_list.rb
./lib/coverband/utils/source_file.rb
./lib/coverband/utils/file_groups.rb
./lib/coverband/utils/lines_classifier.rb
./lib/coverband/collectors/coverage.rb
./lib/coverband/reporters/base.rb
./lib/coverband/reporters/html_report.rb
./lib/coverband/reporters/console_report.rb
./lib/coverband/integrations/background.rb
./lib/coverband/integrations/rack_server_check.rb
./lib/coverband/reporters/web.rb
./lib/coverband/integrations/middleware.rb
./lib/coverband/integrations/resque.rb
./test/coverband/adapters/file_store_test.rb
./test/coverband/adapters/redis_store_test.rb
./test/coverband/at_exit_test.rb
./test/coverband/collectors/coverage_test.rb
./test/coverband/configuration_test.rb
./test/coverband/coverband_test.rb
./test/coverband/integrations/background_test.rb
./test/coverband/integrations/middleware_test.rb
./test/coverband/integrations/rack_server_check_test.rb
./test/coverband/integrations/resque_worker_test.rb
./test/coverband/reporters/base_test.rb
./test/coverband/reporters/console_test.rb
./test/coverband/reporters/html_test.rb
./test/coverband/reporters/web_test.rb
./test/coverband/utils/file_groups_test.rb
./test/coverband/utils/file_list_test.rb
./test/coverband/utils/gem_list_test.rb
./test/coverband/utils/lines_classifier_test.rb
./test/coverband/utils/result_test.rb
./test/coverband/utils/s3_report_test.rb
./test/coverband/utils/source_file_line_test.rb
./test/coverband/utils/source_file_test.rb
./test/integration/full_stack_test.rb
./test/integration/rails_full_stack_test.rb
./test/rails_test_helper.rb
./test/rails5_dummy/config/environment.rb
./test/rails5_dummy/config/application.rb
./test/rails5_dummy/config/coverband.rb
./test/rails5_dummy/config/routes.rb
./test/integration/rails_gems_full_stack_test.rb
./test/fake_app/basic_rack.rb
./test/coverband/integrations/test_resque_job.rb
./test/rails5_dummy/app/controllers/dummy_controller.rb
./test/dog.rb
```
For sure we could work around these test order side effects, but starting with a clean slate feels easier to reason about in the tests. Let's bring this back to real problem that is currently happening on master which is that a test is passing when running the full test suite but not when we run the test in isolation (I have to admit introduced this problem in a recent PR). Running the individual test fails:

```
bundle exec m test/integration/rails_gems_full_stack_test.rb
[Coveralls] Set up the SimpleCov formatter.
[Coveralls] Using SimpleCov's default settings.
Run options: -n "/^(test_this_is_how_gem_it)$/" --seed 8197

# Running:

F

Failure:
RailsGemsFullStackTest#test_this_is_how_gem_it [/Users/karlbaum/workspace/coverband/test/integration/rails_gems_full_stack_test.rb:34]:
Expected nil to be truthy.


bin/rails test /Users/karlbaum/workspace/coverband/test/integration/rails_gems_full_stack_test.rb:27



Finished in 0.621074s, 1.6101 runs/s, 6.4405 assertions/s.
1 runs, 4 assertions, 1 failures, 0 errors, 0 skips
```

Running all the tests passes:

```
➜  coverband git:(master) ✗ bundle exec rake test
/Users/karlbaum/.rubies/ruby-2.5.3/bin/ruby -w -I"lib:lib:test" -I"/Users/karlbaum/.gem/ruby/2.5.3/gems/rake-12.3.2/lib" "/Users/karlbaum/.gem/ruby/2.5.3/gems/rake-12.3.2/lib/rake/rake_test_loader.rb" "test/coverband/adapters/base_test.rb" "test/coverband/adapters/file_store_test.rb" "test/coverband/adapters/redis_store_test.rb" "test/coverband/at_exit_test.rb" "test/coverband/collectors/coverage_test.rb" "test/coverband/configuration_test.rb" "test/coverband/coverband_test.rb" "test/coverband/integrations/background_test.rb" "test/coverband/integrations/middleware_test.rb" "test/coverband/integrations/rack_server_check_test.rb" "test/coverband/integrations/resque_worker_test.rb" "test/coverband/reporters/base_test.rb" "test/coverband/reporters/console_test.rb" "test/coverband/reporters/html_test.rb" "test/coverband/reporters/web_test.rb" "test/coverband/utils/file_groups_test.rb" "test/coverband/utils/file_list_test.rb" "test/coverband/utils/gem_list_test.rb" "test/coverband/utils/lines_classifier_test.rb" "test/coverband/utils/result_test.rb" "test/coverband/utils/s3_report_test.rb" "test/coverband/utils/source_file_line_test.rb" "test/coverband/utils/source_file_test.rb" "test/integration/full_stack_test.rb" "test/integration/rails_full_stack_test.rb" "test/integration/rails_gems_full_stack_test.rb"
[Coveralls] Set up the SimpleCov formatter.
[Coveralls] Using SimpleCov's default settings.
D, [2019-03-12T08:48:01.564119 #24383] DEBUG -- : using default configuration
/Users/karlbaum/workspace/coverband/lib/coverband/at_exit.rb:8: warning: instance variable @at_exit_registered not initialized
/Users/karlbaum/workspace/coverband/lib/coverband/at_exit.rb:10: warning: instance variable @at_exit_registered not initialized
/Users/karlbaum/workspace/coverband/lib/coverband/integrations/background.rb:22: warning: instance variable @thread not initialized
/Users/karlbaum/workspace/coverband/lib/coverband/integrations/background.rb:26: warning: instance variable @thread not initialized
D, [2019-03-12T08:48:01.564412 #24383] DEBUG -- : Coverband: Starting background reporting
/Users/karlbaum/workspace/coverband/test/test_helper.rb:119: warning: method redefined; discarding old test_env
/Users/karlbaum/workspace/coverband/test/coverband/integrations/resque_worker_test.rb:8: warning: assigned but unused variable - queue
/Users/karlbaum/workspace/coverband/test/coverband/reporters/base_test.rb:155: warning: assigned but unused variable - lines_hit
Run options: --seed 31095

# Running:

...................../Users/karlbaum/workspace/coverband/test/fake_app/basic_rack.rb:6: warning: method redefined; discarding old call
/Users/karlbaum/workspace/coverband/test/fake_app/basic_rack.rb:6: warning: previous definition of call was here
.............................................................................................................................................

Finished in 1.941965s, 83.4207 runs/s, 133.8850 assertions/s.
162 runs, 260 assertions, 0 failures, 0 errors, 0 skips
[Coveralls] Outside the CI environment, not sending data.
```

Tried to dig into how not clearing coverage causes the issue but ran out of time on my commute. I did verify however that clearing the coverage like so in setup does ensure that running the individual test and the full test suite expose the same test failure:

```ruby
  def setup
    super
    ::Coverage.result
    # The normal relative directory lookup of coverband won't work for our dummy rails project
```

I am confident that we will dig into this specific issue and get to the bottom of it but I think it's a great example of how sharing state between tests can cause confusing results.

danmayer:
Thanks that is really helpful. Let me think a bit more about it and see if I can come up with anything... 

Currently, one thought would be we run tests twice, once clearing coverage, to ensure all tests run in isolation. Then we run again without that and record test coverage of actual Coverband code. This means one would have to write tests that work fully in isolation, but would also expect to work with the shared data, which is a little odd / confusing, the empty example test wouldn't work for example but might give us much of both benefits.

kbaum:
The running test twice sounds very interesting. Also wonder if there is a way to persist the coverage after each test and then combine all of the results. Kind of like how one would generate coverage when running parallel tests in separate ruby processes.

danmayer:
Oh, that is a great idea as well.

kbaum:
Closing this for now as i started experimenting with forking each test.

PR #207, opened 2019-03-13
Make slight refactor on array_diff

kbaum:
Silly and small but was just reviewing your last commit and now I get the weird problems forking was causing.


danmayer:
wow nice this is awesome I like it @kbaum merging

PR #213, opened 2019-03-16
Add require_unique_files

kbaum:
Allow for requiring unique "dog.rb" files for testing isolation. Idea is that when calling require_unique_file, a unique file is generated using SecureRandom.uuid as part of the path. After all tests have run, the unique files are removed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ temp_results
 .byebug_history
 .env
 log/
+test/unique_files
+test/rails5_dummy/tmp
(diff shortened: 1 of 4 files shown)


danmayer:
cool this will be interesting in trying to reduce test dependency issues. I was originally thinking of using Ruby's tmp file but honestly, I don't think it adds anything over the method you took... https://ruby-doc.org/stdlib-1.9.3/libdoc/tempfile/rdoc/Tempfile.html

kbaum:
Yeah originally went down the temp_file route but wasn't able to make it work. Can't quite remember why but I think requiring a TempFile was giving me difficulty.

PR #214, opened 2019-03-17
Store eager load data in separate key

kbaum:
While rails is initializing, coverage data is stored under an eager_loading key. With this change all should work the same as coverage report generation pulls and merges from both keys.

An attempt to address #186. The next step would be to modify reports to display both eager and runtime coverage data.


test/integration/rails_gems_full_stack_test.rb
```diff
@@ -1,36 +0,0 @@
-# frozen_string_literal: true
```
kbaum: @danmayer Was running into trouble having both rails_gem_full_stack_test and rails_full_stack_test as rails can only be initialized once per test run.

danmayer:
Interesting looks like a good approach @kbaum let me look / think a bit more. I was originally thinking of something less tied to Rails like the first resport would count as load and everything else would be runtime, this is likely a better path, but we will need to make a very clear and easy API for Sinatra / Non-Rails...

* How would this be impacted for Resque / Sidekiq, they always just report runtime, and we relying on the Rails boot to capture code loading?
* For Coverage reports, other than grabbing the different time stamps in the report, do you think colors or line annotations would be helpful? I am thinking a bit more about what data we might need to capture to drive the UI changes.

kbaum:
Gotcha. 

> I was originally thinking of something less tied to Rails

Also was aiming to make this not specific to rails. The railtie calls `eager_loading_coverage!` and `runtime_coverage!` but these could just as easily be called at the right points by any other type of application.

> How would this be impacted for Resque / Sidekiq, they always just report runtime, and we relying on the Rails boot to capture code loading?

I believe rails apps that use sidekiq and resque load the entire rails app on boot. Would think they also eager load the application and the railtie hooks should fire in the same way. Will test but I don't think it would be handled differently.

> For Coverage reports, other than grabbing the different time stamps in the report, do you think colors or line annotations would be helpful?

Good question. I think the most important goal is to be able to see runtime coverage without eager load coverage. One idea could be just adding a "% Runtime covered"

image

Or we could add "runtime" and "eager load" checkboxes but that would require much more work:

image

I think a different color representation for "eager loaded" lines would be helpful as well. I would imagine users will want to dive in and see the detail.

danmayer:
cool thanks @kbaum since we resolved a bug on info, resque memory, and the negative line numbers without changing any of the APIs... I am thinking we release 4.1.1 with those minor fixes. Then we can merge this PR (or just keep building on this branch) and start on the UI work. I think the UI suggestion about breaking out total coverage and runtime coverage is the biggest helper and then on the individual code view being able to distinguish eager load and runtime. We can start to work on that from this.

kbaum:
> I am thinking we release 4.1.1 with those minor fixes

Agreed

> Then we can merge this PR (or just keep building on this branch) and start on the UI work. 

I tried to make this one backwards compatible with what's on master. We could merge it and then we would have the benefit of it being used in real world scenarios. The only downside is there is no real feature benefit as of now. Was thinking an mvp might be allowing for rake coverage generation with just runtime data. Possibly with an environment variable? Something like:

```
COVERAGE_TYPE=runtime rake coverband:coverage
```

danmayer:
I think given some risk and that it will increase the memory used in redis, it would be good safe this for the next release. We can do additional testing and put it through some paces as well as surface some functionality driven by it. I can work on the UI updates over the next couple days and we can cut a beta release to test on some projects very soon.

danmayer:
OK, released 4.1.1. Really happy to finally have whatever was causing the Resque memory leak resolved, wish I fully understood it but happy enough to have everything functioning and getting exercised on some high traffic apps.

kbaum:
> I think given some risk and that it will increase the memory used in redis, it would be good safe this for the next release. 

Completely agree that this should be a next release thing. I was thinking that we could merge into master once we've cut a release but I'm also down with working off a branch.

danmayer:
Ah, that makes sense and works for me. Let's merge this now that it is released as I agree it is much easier to look at and review smaller PRs.

PR #218, opened 2019-03-26
Reset collector on configure

kbaum:
On master the following was failing:

```
bundle exec m test/integration/rails_full_stack_test.rb
```

The problem was that even though we were calling Coverband#configure with track_gems set to true within the test setup, Collectors::Coverage @track_gems had already been set to false. This is most likely only a problem in tests as Coverband#configure is invoked more than once.


danmayer:
nice thanks

Issue #161, opened 2018-12-30
Coverage last activity recorded not acting as expected


danmayer:
Coverage last activity recorded is always the most recent coverage collection, not when that file was hit.

The new feature that displays when coverage timeframe (first seen for the current version of a file and when it last received an update)... Always, has the last updated at the most recent time Coverband reported...

https://coverband-demo.herokuapp.com/coverage?#706772eba501dc088a894962fcc23c960652ac34

* I think we want to not report any updates when there are 0 new lines hits
* I think on app boot it will always update just because it counts as a hit for method definitions and such
         * can we filter those out perhaps? or not update the timestamp from that boot coverage report?

danmayer:
this and #186 are kind of the same and will likely be looked at for version 4.3.0

danmayer:
targetting 4.2.0 for this fix

danmayer:
closing as this is on master and the 4.2.0.rc2 release

PR #219, opened 2019-03-26
WIP to utilize the eager load vs runtime coverage

danmayer:
OK, at least got a branch going with some initial work...

* This WIP actually already supports a feature, woohoo, the PR fully fixes issue #161 

An actively used controller showing last hit activity
Screen Shot 2019-03-25 at 9 52 16 PM

An unreachable model showing no last activity
Screen Shot 2019-03-25 at 9 52 29 PM

* I think this exposes a bit of an overly tight coupling in the code
        * now we have reports, coverage, all aware of the multiple storage types, and sometimes getting single reports or merged
        * I think I want to refactor so that only the store is aware of the types... so you don't call merged_reports from the report you just call `coverage_data` which would fetch the underlying types formatted as needed... The reports of course will now need all the data, so they will understand the difference between runtime, eager_load, and merged but they won't have to ask for multiple types from the store
         * I am going to try to refactor this to make what I am talking about more clear
* I updated the web debug to be way more useful now showing all 3 coverage data sets eager, runtime, and merged.


lib/coverband/adapters/base.rb
```diff
@@ -76,10 +86,11 @@ def file_hash(file)
       def expand_report(report)
         expanded = {}
         report_time = Time.now.to_i
+        updated_time = self.type == Coverband::Collectors::Coverage::EAGER_TYPE ? nil : report_time
```
kbaum: This is great! I didn't think of the benefit we would get here with tracking last report time.

lib/coverband/adapters/redis_store.rb
```diff
@@ -22,7 +22,13 @@ def initialize(redis, opts = {})
       end
 
       def clear!
```
kbaum: I think I introduced a thread safety issue here in that we only have one instance of our store per ruby process. If two different threads call `clear!` at the same time we could have unintended results since we are depending on the the `type` state of the store.

I had started by passing the type as a method argument but it required a lot of passing that type state from method to method and didn't feel right. 

We could solve this by either creating a separate store per thread or introducing a Mutex. I think introducing a separate store per thread might be overkill considering then we would probably want a separate redis connection per thread?

lib/coverband/collectors/coverage.rb
```diff
@@ -10,6 +10,10 @@ module Collectors
     # and this method is currently in a ensure block in middleware and threads
     ###
     class Coverage
+      RUNTIME_TYPE = nil
```
kbaum: This is a great change as it makes the code more readable. I felt a little funny about making this nil honestly but it gave backwards compatibility to existing data so felt it was worth it.

kbaum:
Great idea for a first feature to utilize eager_loading coverage!

lib/coverband/adapters/base.rb
danmayer: yeah I was really happy to see how easy it was to implement this after you change, makes me think you nailed the breakdown on the data we need. Awesome!

lib/coverband/adapters/redis_store.rb
danmayer: yeah I want to refactor the way the store works with 2 types, good point on the thread issue. When I refactor it to only be a concept known about by the store I will try to address.

lib/coverband/collectors/coverage.rb
danmayer: yeah, I have been considering this... I might actually change it to `:run_time` and now that we have built migrations, have the migration upgrade the `nil` stored value to `runtime`... can't decide entirely. I was going to get everything working and then rethink the value of migrations etc.

lib/coverband/adapters/redis_store.rb
danmayer: I haven't resolved this yet

danmayer:
OK, I don't love this yet... I think I can simplify access to all the different data modes... Also, this feels like it is probably horrible in terms of memory and perf for rendering the data... but that won't impact runtime performance just the reporting... Take a look @kbaum 

I broke a bunch of tests but this let me start to play with the APIs of how data access would work...

I am not sure if the whole `method_missing` is a good idea long term, but it sure was a quick way to allow deferring to any of the result types.

I haven't built out all the new metrics but have shown on the index and source file pages how the important data can now be accessed... Now a bit more thought on UI and what data is really helpful.

Screen Shot 2019-03-30 at 3 49 30 PM

Screen Shot 2019-03-30 at 4 08 40 PM

lib/coverband/utils/coverband_results.rb
```diff
@@ -0,0 +1,45 @@
...
+module Coverband
+  module Utils
+    class CoverbandResults
+      attr_accessor :type, :results
+
+      def initialize(report)
+        self.results = {
+          Coverband::EAGER_TYPE => Coverband::Utils::Result.new(report[Coverband::EAGER_TYPE]),
+          Coverband::RUNTIME_TYPE => Coverband::Utils::Result.new(report[Coverband::RUNTIME_TYPE]),
+          Coverband::MERGED_TYPE => Coverband::Utils::Result.new(report[Coverband::MERGED_TYPE])
```
kbaum: RE the concern about memory. If `self.results` was of a different type, could have merged created only when needed? Lazily?

views/file_list.erb
```diff
@@ -37,9 +39,15 @@
             <%= link_to_source_file(source_file) %>
           </td>
           <td class="<%= coverage_css_class(source_file.covered_percent) %> strong"><%= source_file.covered_percent.round(2).to_s %> %</td>
+          <!-- <td class="<%= "#{coverage_css_class(source_file.covered_percent)}_runtime" %> strong"><%= source_file.covered_percent.round(2).to_s %> %</td> -->
           <td><%= source_file.lines.count %></td>
           <td><%= source_file.covered_lines.count + source_file.missed_lines.count %></td>
           <td><%= source_file.covered_lines.count %></td>
+          <td>
+            <%= result.from_type(Coverband::RUNTIME_TYPE) do
```
kbaum: Probably an opportunity to introduce a method into `CoverbandResults` to do this?
```erb
<%= result.covered_lines(source_file.filename, type: Coverband::RUNTIME_TYPE) %>
```

Although it would feel more consistent to just have that on source_file.

```erb
<%= source_file.covered_lines(type: Coverband::RUNTIME_TYPE).count %>
```

views/layout.erb
```diff
@@ -33,7 +33,7 @@
       <ul class="group_tabs"></ul>
 
       <div id="content">
-        <%= formatted_file_list("All Files", result.source_files) unless view_gems? %>
+        <%= formatted_file_list("All Files", result, result.source_files) unless view_gems? %>
```
kbaum: Minor point but is an opportunity to cut down the number of method params here by invoking `result.source_files` within `formatted_file_list` method?

```erb
<%= formatted_file_list("All Files", result) unless view_gems?
```

kbaum:
This looks great! I love the breakdown of load, runtime, and all in the report. Thinking it might also make sense to introduce a new color for only load coverage? Not sure.

lib/coverband/utils/coverband_results.rb
danmayer: yeah lazy processing could be interesting, I guess I will see how bad memory and perf is on larger projects

views/file_list.erb
danmayer: now like this `<%= result.file_with_type(source_file, Coverband::RUNTIME_TYPE).try(:covered_lines).try(:count) || 0 %>` which is better than it was, I don't love needing to use `try` as not all files or lines will exist in runtime coverage.

views/layout.erb
danmayer: good call cleaned this up

danmayer:
Screen Shot 2019-04-07 at 3 39 16 PM

I think this might be good enough for the first version of this. I will work to clean up the code and get tests so we can get it merged. I am thinking a bigger rewrite of the front end HTML generation stuff will be in order but that could follow.

lib/coverband/adapters/redis_store.rb
danmayer: resolved the race conditions

lib/coverband/adapters/base.rb
```diff
@@ -57,22 +57,14 @@ def covered_lines_for_file(file)
       protected
 
       def split_coverage(types)
-        original_type = type
         types.reduce({}) do |data, type|
```
kbaum: Avoid setting any state by allowing the method param on get_report. Great solution!

views/file_list.erb
```diff
@@ -37,9 +39,16 @@
             <%= link_to_source_file(source_file) %>
           </td>
           <td class="<%= coverage_css_class(source_file.covered_percent) %> strong"><%= source_file.covered_percent.round(2).to_s %> %</td>
+          <% runtime_percentage = result.file_with_type(source_file, Coverband::RUNTIME_TYPE).try(:covered_percent).try(:round, 2) %>
```
kbaum: This looks much better. Not important but could move more logic into result object?

```erb
<%  = result.covered_percentage(source_file, Coverband::RUNTIME_TYPE) %>
```

lib/coverband/utils/coverband_results.rb
```diff
@@ -0,0 +1,42 @@
...
+####
+# A way to access the various coverage data breakdowns
+####
+module Coverband
+  module Utils
+    class CoverbandResults
+      attr_accessor :type, :results
+
+      def initialize(report)
+        self.results = {
```
kbaum: Minor suggestion.. holds up if we had more types i think.
```
self.results = Coverband::Types.each_with_object({}) do |type, hash|
  hash[type] = Coverband::Utils::Result.new(report[type])
end
```

views/file_list.erb
danmayer: yeah I thought about that a bit, the problem is that the result object is pretty dumb in that it is just calling `method_missing`... I could make assumptions like the last argument is the type chop it off and use it to select the type and then call with the rest of the args, but that seemed a bit more complicated and magical, than the above approach.

lib/coverband/utils/coverband_results.rb
danmayer: good call

danmayer: I will update and do another pass on tests tonight

danmayer: ok resolved

danmayer:
OK, @kbaum I want to knock this out over the weekend remaining questions...

* Spend time figuring out a good DB migration?
* I need to run some more perf checks as I think this could be really bad on large projects
    * if so I might need to take your lazy loading suggestion and move file details to ajax

danmayer:
Actually one more issue it looks like by default the first hit on heroku counts as runtime not load time...

https://coverband-demo.herokuapp.com/coverage/#_App

so debugging that and perf seems like the blocker for the weekend.

kbaum:
> if so I might need to take your lazy loading suggestion and move file details to ajax
I guess if we start using ajax for file details, we would have to stop supporting the static building of the report from the rake task?

`Actually one more issue it looks like by default the first hit on heroku counts as runtime not load time...`

That is surprising. Probably a silly question, but did you clear the previous redis data?

kbaum:
> Spend time figuring out a good DB migration?

I wasn't thinking there would be a migration for this since we are just storing the eager load coverage data under a different redis key. In order to get accurate eager load data vs runtime data, was thinking clearing all coverage data would be needed.

kbaum:
> Probably a silly question, but did you clear the previous redis data?
Never mind, I am seeing the same thing.

danmayer:
Yeah we could avoid the migration and just warn people the new functionality wouldn't work without a reset.

The heroku issue is odd and I wasn't expecting that, think I will perhaps work on the ajaxified partials today and look at that tomorrow. If you have any ideas @kbaum let me know

kbaum:
Would love to help more but unfortunately wont be at computer for the next week as off on a spring break road trip vacation with family.

> think I will perhaps work on the ajaxified partials today and look at that tomorrow.

Now that I am at a company with an api only rails app, I am more aware of the need for coverband working with a rake task or uploading to s3. I wonder if we could still use ajax but have the rake task just download the json locally? Then could start some type of server locally to serve the json.

kbaum:
Just doing some debugging of this locally with the coverband_demo and it seems to be working. I created a post and i see this in the coverage report for posts_controller:

image

All the runtime and load annotations seem to be correct. Do you only see this issue on heroku?

danmayer:
Yeah heroku only which is weird

kbaum:
Very weird. Is eager loading enabled for that environment?

danmayer:
yeah it does eager load, but I have eager loading on in dev as well to try to ensure it is a similar environment while testing.

prod
https://github.com/danmayer/coverband_demo/blob/master/config/environments/production.rb#L11

dev
https://github.com/danmayer/coverband_demo/blob/master/config/environments/development.rb#L11

danmayer:
figured it out, which took way to long... `RAILS_ENV=production bundle exec rake assets:precompile` sends runtime data from building the assets. Working on a default work around that excludes that, this is kind of a heroku only thing, not sure many people build assets prior to deploy, but on different systems... turning off at_exit support resolves...

kbaum:
Great catch! I would have thought the railtie would kick in even in assets:precompile and it would have recorded in the eager_loading bucket. My guess is that assets precompile disabled eager load?

I’m honestly not sure how other platforms run assets:precompile as I’ve been on heroku for years. Seems like it could be a problem elsewhere though depending on where assets are precompiled.

danmayer:
hmmm while asset building locally and heroku runs two tasks... `['assets:clean', 'assets:precompile']1` ignoring those or even entirely disabling the at_exit still seems to show 'loading' coverage ending up in the runtime store ;(

Now adding a bunch of debug logs to see if I can figure out what is happening.

kbaum:
think background thread could be causing this?

danmayer:
Yeah had to stop for the day but just noticed background thread is
triggered on asset builds too... So digging in more that direction, tomorrow

danmayer:
OK well, that was no fun... but I guess I understand and have it working. I need to clean up a bit more, fix broken tests, and add some tests around what all I was debugging.

Screen Shot 2019-04-14 at 2 45 37 PM

In the end, I basically set up to ignore the first hit while in runtime mode. This does throw away some data in some cases, but it protects against all sorts of hard to predict scenarios... Things like new process / thread that creates a Coverband Collector instance which has no previous_coverage data... Which means it will send ALL coverage data not just the runtime data.

1. rake tests were an issue, at_exit while handled by rail ties also loads by default
       * specific rake tasks now ignored, but that doesn't matter because of the 'runtime' ignore first hit
2. yes background was getting a trigger on some things we didn't expect
3. The server check was failing for me locally
4. The local behavior changed for me after merging in master
    * this seems to be related to resetting coverage on configure: https://github.com/danmayer/coverband/pull/218/files#diff-a637af9972be5b34783d41387d25e0bfR50
     * I think that is OK and my new solution works for that, but this was somehow getting hit 5+ times on boot
     * I think it was hit once per puma thread, but didn't dig too deep

kbaum:
Seems like the first implementation I did was quite brittle. Ignoring the first runtime report is a clever solution!

In my head, as long as the one off processes eagerly initialized rails, then we wouldn’t have run into a problem with class loading coverage data being reported as runtime data. Were the problems all related to processes not eagerly initializing rails? Am I not thinking about it right?

danmayer:
your thinking is right and I really am struggling to fully understand it. What I believe is happening is that the server starts up and records the initial eager_loading coverage as we expect... Then if I don't hit the server it never sends more coverage, this makes sense because the background job is started by the first hit to the rack middleware. 

The issue is when it then goes to save coverage the previous_results are nil, but we are now in runtime mode, so it will report all the first set of data in the Coverage object, not what is just new...

This means between the `after_initialize` the previous_results is not passed over... This seems to be because when rack middleware triggers background to call `Coverage.instance` it is on a new thread.

My thinking is eager_loading runs before puma forks off and therefor all the first runtime hits are in new threads.

So it makes sense that the first runtime hit has the FULL std lib Coverage report and it doesn't have the in memory diff to clean it up. ;(

So yeah I think I keep the ignore first hit in runtime mode if previous_results is nil. Which feels a bit hacky but also seems like a robust way to prevent this issue.

kbaum:
Forking of processes often causes confusion for us. Ignoring first runtime hit is a robust solution. The only downside could be missing coverage data in short lived ruby script runs. 

Would be interested in trying to reproduce the problem. I tried a [few days ago](https://github.com/danmayer/coverband/pull/219#issuecomment-482813705) but came up short. Back then it sounds like we were only able to reproduce on heroku. Still the case?

> This means between the after_initialize the previous_results is not passed over... This seems to be because when rack middleware triggers background to call Coverage.instance it is on a new thread.

It looks like previous_results is a class variable so I would think all threads would share this. Perhaps you meant Coverage.instance is on a new process?

lib/coverband/at_exit.rb
```diff
@@ -5,14 +5,30 @@ class AtExit
...
+
+          #####
+          # TODO: This is is brittle and not a great solution to avoid deploy time
+          # actions polluting the 'runtime' metrics
+          #
+          # * should we skip /bin/rails webpacker:compile ?
+          # * Perhaps detect heroku deployment ENV var opposed to tasks?
+          #####
+          default_heroku_tasks = ['assets:clean', 'assets:precompile']
+          if defined?(Rake) && Rake.application.top_level_tasks.any? { |task| default_heroku_tasks.include?(task) }
```
kbaum: Getting this error when I exit from puma on coverband_demo:
```
/Users/karlbaum/workspace/coverband/lib/coverband/at_exit.rb:26:in `block (2 levels) in register': undefined method `application' for Rake:Module (NoMethodError)
```

danmayer: cool fixed, will push soon

danmayer:
So I was able to reproduce locally after pulling in master. I first reproduced via the `rake asset tasks` but after messing around a bit I was able to have this issue occur locally.

> It looks like previous_results is a class variable so I would think all threads would share this. Perhaps you meant Coverage.instance is on a new process?

It is a class variable, but when you are on a thread and call Coverband.instance it will create a new object via the private method, which then calls `reset_instance` which sets the class var to nil via `  @@previous_results = nil`... Now I thought this could just be related to that, but it appeared that @previous_results was already nil in some cases...

So I think per thread calling reset_instance messes up thread based servers... Then the losing the class var on process fork causes issues for process-based servers. 

I would like to find a better way that doesn't lose any data, but I think this brief gap throwing away some initial data might be the best first version.

kbaum:
I was able to reproduce by setting workers to 1 within puma config:

https://gist.github.com/9f28ca1cc1b94262e516e0f1643b5f5b

I feel like we have an inconsistency in that we have coverband instances per thread that share a class level per process `@@previous_results` variable. I am thinking makes sense to change coverband from an instance per thread to an instance per process. This would be more consistent with the way coverband has evolved in that we are using the ruby std coverage library which also stores coverage data globally per process. Also now that we are using one background thread to report to redis, we should not need a coverband instance for each thread.

danmayer:
yeah we left the thread version so folks could use it as per rack middleware, but that isn't the default and likely no one is using it. The issue between thread and process variable is confusing at the very least. Although in this case we aren't losing it at the thread level but the process level. I will get rid of the threading tomorrow and see if that helps and simplifies things a bit. 

* drop threads
* ajax load file data, to resolve perf slowdown
* possible get in the feature to delete coverage per file, I got another request for that feature today.

My goal is to have this wrapped up for a release by Monday.

danmayer:
OK, so I think all that is left is working on the ajax load vs rendering all file data, I will try to work on that a bit later.

danmayer:
OK, I got the ajax loader working as well, this should help a lot for larger apps.

You can check it out on the demo site: https://coverband-demo.herokuapp.com/coverage#_App

I think tomorrow I can clean up a bit, add some test coverage, merge everything along with your PRs and put it on a few sites to test a release candidate... so we can hopefully publish the gem on Monday or Tuesday...

I added notes to work on lazy loading for next version, I think we can also look at getting robucop / code climate going, with related cleanup to get everything matching rules, and improving the situation around building the report locally which you mentioned... let me know if you have any other thoughts @kbaum

danmayer:
Also, this is now 10K LOC because I added a JS formatter to clean up the JS but it expanded compressed jquery, I will try to split our JS to compressed dependancies and app code tomorrow.

danmayer:
combined with the ajax the lazy loading will be a huge win, right now to render the single file info it builds the entire Coverband Results data

danmayer:
fixed the JS expansion issue, and improved the loader

danmayer:
ok @kbaum I have this running with the published RC2 gem in production on high traffic apps, if all looks good today I will make the official release tomorrow. If you run it on anything or see issues let me know

kbaum:
Great! Plan on pushing this live sometime today. Will let you know.

danmayer:
@kbaum I am debugging an issue where runtime coverage is showing up on method definitions in production.  I have track down at least one way this can pollute the runtime coverage is from resque workers, so I am going to debug and fix that, likely will need another release candidate and possibly another day or so.

kbaum:
@danmayer - this may be a silly question but is the `rake coverband:coverage` task still expected to work now that we are using ajax loading? Tried to run it and it just exited with no error message and didn't open the report.

danmayer:
ouch, I totally forgot that! It would be easy to generate the state version when not using the web. That is a good bug, can you file an issue I can fix that. Need some additional testing around the split html reporting views, static vs dynamic.

danmayer:
Sorry about that @kbaum

kbaum:
Sounds good. Will open an issue. Thanks @danmayer!

PR #224, opened 2019-04-02
Fix COVERBAND_DISABLE_AUTO_START not working for rails

kbaum:
Fixes #224


danmayer:
I think this looks good only way I can think to test would be to shell a process that starts rails then check the redis after shutting it down.

danmayer:
Feel free to merge if you can't figure out a good way to test @kbaum

kbaum:
I think i'll merge for now. I think we will eventually need to come up with a forking strategy for testing our rails integration under multiple scenarios.

PR #225, opened 2019-04-06
Check if thread is still alive

kbaum:
When a process forks, the thread instance variable will be copied to the new process but the thread will not be running. This checks the thread to see if if it is alive. If it is not, a new thread is started.

Noticed this issue when trying to get coverband to work with https://github.com/jondot/sneakers. Tried to start the background thread after_fork:

```ruby
  after_fork: lambda do
    Coverband::Background.start
  end
```
This did not work since the `@thread` instance was not null but the thread itself was no longer running.

### DIFF
--- a/lib/coverband/integrations/background.rb
+++ b/lib/coverband/integrations/background.rb
@@ -15,19 +15,20 @@ module Coverband
     end
 
     def self.running?
-      !!@thread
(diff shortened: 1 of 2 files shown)


danmayer:
Nice, this makes sense.

kbaum:
Thanks!

PR #227, opened 2019-04-09
Remove puts statement

kbaum:


kbaum:
@danmayer Assuming this one is ok to just merge as i added this puts statement on accident.

danmayer:
yep cool thanks

PR #228, opened 2019-04-20
Extract delta coverage class

kbaum:
Small refactor to extract coverage delta logic into separate class.

### DIFF
--- a/lib/coverband/collectors/coverage.rb
+++ b/lib/coverband/collectors/coverage.rb
@@ -1,6 +1,4 @@
 # frozen_string_literal: true
-require 'forwardable'
-
 require 'singleton'
 
 module Coverband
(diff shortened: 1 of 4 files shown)


danmayer:
Nice this is great, thanks @kbaum 

I think it might conflict with some work on the current branch could you fetch that branch and ensure it works and have the PR target that as the base branch? This does look cleaner, was there any impact or change on the branch marks... this used to be part of the really tight event loop but since it is in a background thread I wouldn't expect many issues.

danmayer:
I guess since we aren't releasing I should have got the long-running feature branch on master sooner, just kept hitting various issues making that messier than I have liked it.

kbaum:
> I think it might conflict with some work on the current branch could you fetch that branch and ensure it works and have the PR target that as the base branch?

Makes sense. Will do.

> was there any impact or change on the branch marks

Will check. Don't think this changes much of the logic so I think we should be good.

lib/coverband/collectors/coverage.rb
```diff
@@ -82,27 +79,16 @@ def track_file?(file)
 
       private
 
-      def add_filtered_files(new_results)
-        new_results.each_pair do |file, line_counts|
-          next unless track_file?(file)
-          add_file(file, line_counts)
-        end
+      def filtered_files(new_results)
```
kbaum: @danmayer avoid the use of the instance variable here by just doing it all in one method that returns the filtered files.

kbaum:
@danmayer rebased with `feature/runtime_coverage_ui`.

lib/coverband/collectors/coverage.rb
danmayer: nice a little harder to mentally unpack but really concise and likely better memory / perf

danmayer:
nice thanks @kbaum this turned out great!

PR #229, opened 2019-04-20
Test report coverage verbose exception

kbaum:


lib/coverband/collectors/coverage.rb
```diff
@@ -138,9 +138,7 @@ def initialize
         if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.3.0')
           raise NotImplementedError, 'not supported until Ruby 2.3.0 and later'
         end
-        unless defined?(::Coverage)
```
danmayer: I can't recall exactly but I thought I added this because it was given a warning when already loaded by things like simplecov, etc

danmayer:
looks good, can you send the PR to the current feature branch, hoping to get everything merged up tomorrow

lib/coverband/collectors/coverage.rb
kbaum: Interesting.. think it's ok to remove for now and add back in if needed?

kbaum:
Set PR to target `feature/runtime_coverage_ui`.

lib/coverband/collectors/coverage.rb
danmayer: yeah should be OK

PR #232, opened 2019-04-22
fix for resque boot storing runtime coverage

danmayer:
OK, this resolves issues where resque booting would store boot time coverage as runtime coverage.

I think this brings up if this would be an issue with other integrations, or if we need to better document how to integrate this for things like default Rack apps.

take a look @kbaum and I will look at doing a bit more testing and try to see if this is the only issue

### DIFF
--- a/lib/coverband/integrations/resque.rb
+++ b/lib/coverband/integrations/resque.rb
@@ -2,9 +2,11 @@
 
 Resque.after_fork do |job|
   Coverband.start
+  Coverband.runtime_coverage!
 end
 
 Resque.before_first_fork do
+  Coverband.eager_loading_coverage!
   Coverband.configuration.background_reporting_enabled = false
   Coverband::Background.stop
   Coverband::Collectors::Coverage.instance.report_coverage(true)
--- a/lib/coverband/integrations/resque.rb
+++ b/lib/coverband/integrations/resque.rb
@@ -3,6 +3,8 @@
 Resque.after_fork do |job|
   Coverband.start
   Coverband.runtime_coverage!
+  # no reason to miss coverage on a first resque job
+  Coverband::Collectors::Delta.set_default_results
 end
 
 Resque.before_first_fork do
--- a/lib/coverband/collectors/coverage.rb
+++ b/lib/coverband/collectors/coverage.rb
@@ -5,7 +5,6 @@ require_relative 'delta'
 module Coverband
   module Collectors
     ###
-    # TODO: look at alternatives to semaphore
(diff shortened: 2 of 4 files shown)


lib/coverband/integrations/resque.rb
```diff
@@ -2,9 +2,11 @@
 
 Resque.after_fork do |job|
   Coverband.start
+  Coverband.runtime_coverage!
 end
 
 Resque.before_first_fork do
+  Coverband.eager_loading_coverage!
```
kbaum: Interesting, I had thought the railtie would handle this if we are running resque within rails.

danmayer: I am fixing the test, but resque workers don't seem to call the full rails initialization, I will dig a bit more into why the rails hooks aren't taking care of this... but in our setup without this change while it does seem to preload our app code it sends a first hit before running any job with coverage in the runtime state.

danmayer: also, I believe sidekiq is fine based on how it loads

kbaum: Oh. Good to know about sidekiq. Guess we’ll have to handle on a case by case basis.

danmayer: OK, so I believe the issue comes down to various ways to run resque but a common way is

`bundle exec rake environment resque:work`

This runs `rake environment` which boots the full rails app and then runs resque, but while our rails hooks fire, the files aren't loaded this is because of a Rails perf hack, `RAKE ignores eager_load`

```
# Rake tasks automatically ignore this option for performance.
config.eager_load = true
```
https://github.com/rails/rails/blob/df8ee09ce71338cdf9816225df1bdebc707f3560/actiontext/test/dummy/config/environments/production.rb#L12`

Resque in its task LATER forces eager_loading if the rails config option is set, but this happens AFTER our rails hooks have fired and set coverband back into runtime mode assuming that code loading is all done.

This took me far to long to figure out as did figuring out how to fix the resque test.

danmayer:
take a look and let me know if everything makes sense and in the morning I can make RC3 and release that and test it in production.

lib/coverband/integrations/resque.rb
kbaum: Great work! Was looking into how they disable eager load and it looks like they just set it to false within the environment rake task.

https://github.com/rails/rails/commit/9cac69c602f57e397fe01d866cb24ce1781606d4#diff-580e183141be706162224ef696bdb6c7R294

PR #236, opened 2019-04-25
Use rack for coverband:coverage

kbaum:
This was easier than I thought. We could also just include this as a separate rake task to make everything backwards compatible but it feels tempting to just support the coverage report through rack. As the report gets more advanced it will be difficult to support both.

### DIFF
--- a/lib/coverband/utils/tasks.rb
+++ b/lib/coverband/utils/tasks.rb
@@ -17,6 +17,12 @@ namespace :coverband do
     end
   end
 
+  desc 'report runtime Coverband code coverage'
+  task :coverage_server do
+    environment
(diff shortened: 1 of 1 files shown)


danmayer:
Nice this is cool and much simpler than I would have guessed. The issue I see is that the points you made about your current app not being able to support web... In that case this would still work for you because you can access the production Redis... My last and current gig no longer allow access to production Redis... IF we go purely this route we would not support the CI flow anymore, which was really nice for me for a good number of years.

Kind of leaning towards supporting a static and dynamic report... because of that, as things like GDPR are going to continue to make production access less and less common.

danmayer:
Of course, the further alternative is a hosted service, that pulls files through git for source views ;)

kbaum:
@danmayer moved to new rake task. Let me know what you think.

danmayer:
awesome this is really cool, thanks @kbaum

PR #238, opened 2019-04-27
Prepend eager load

kbaum:
This approach prepends to the `Rails::Engine#eager_load!` method. The reason I went down this path is because in a project that uses sneakers, the eager_load! was explicitly being called.

https://github.com/jondot/sneakers/wiki/Eager-loading

I believe using this approach will avoid the issues we were running into with rails rake tasks disabling eager_load: 

https://github.com/rails/rails/commit/9cac69c602f57e397fe01d866cb24ce1781606d4#diff-580e183141be706162224ef696bdb6c7R294

Using this approach, the eager loading logic only comes into play when eager_load is actually being used. We should not need #233 anymore.

### DIFF
--- a/lib/coverband/collectors/coverage.rb
+++ b/lib/coverband/collectors/coverage.rb
@@ -46,10 +46,7 @@ module Coverband
         # when we are in runtime collection mode, which do not have a cache of previous
         # coverage to remove the initial stdlib Coverage loading data
         ###
-        if ((original_previous_set.nil? && @store.type == Coverband::EAGER_TYPE) ||
-           (original_previous_set && @store.type != Coverband::EAGER_TYPE))
-          @store.save_report(files_with_line_usage)
-        end
+        @store.save_report(files_with_line_usage)
       rescue StandardError => err
         if @verbose
           @logger&.error 'coverage failed to store'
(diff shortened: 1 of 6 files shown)


danmayer:
oops meant to close the other PR and merge this one... fixing

danmayer:
anyways I like this approach, very nice.

kbaum:
Great! I feel like I am able to better understand our integration now. We are reporting eager_loading with rails only when eager_load is being used. 

Thinking about rake tasks now. When a class gets loaded from a rails rake task now, it will be recorded as runtime coverage. I think that makes sense though given it is happening outside of an eager_load.

danmayer:
> Thinking about rake tasks now. When a class gets loaded from a rails rake task now, it will be recorded as runtime coverage. I think that makes sense though given it is happening outside of an eager_load.

I mostly agree and as we are discussing in our current PR, I believe the issue is deploy tasks run in production mode as they end up loading all the models/controllers etc they would load all that as runtime and that would basically break the value of the eager loading right?

kbaum:
Agreed. If `rake asset:compile` and/or `rake db:migrate` are loading up all of the models/controllers it would definitely defeat the purpose of eager_loading. Was thinking that maybe we have this labeled wrong. Instead of `eager_loading` perhaps should we call it `initialization`? Then for the rails integration, we could capture all of the rails app initialization and eager_loading into the initialization bucket. The term `initialization` could mean different things for different frameworks but there's always going to be an initialization vs runtime for any process. Loading the models/controllers from rake tasks would fit under initialization.

PR #239, opened 2019-04-28
Remove previous_coverage code

kbaum:
Since we now explicitly capture eager_loading coverage within `Rails::Engine#eager_load!`, we no longer need this logic.

This is a continuation of #238


danmayer:
sorry some of this changed in my rubocop branch which I merged, should be easy to figure out but likely easier for you since you authored original and this change, can you fix up the conflict

danmayer:
@kbaum

kbaum:
@danmayer Fixed conflicts.

PR #243, opened 2019-05-01
Move semaphore lock up the stack

kbaum:
Coverage#report_coverage should never be called by two threads at a time.


danmayer:
I was waiting for you to fix the tests, but I see now they are the bundler issue, fixing that and will merge that to master and when tests pass this looks good

Issue #234, opened 2019-04-24
Rake coverband:coverage task no longer working


kbaum:
Version: 4.2.0.rc2

Run `rake coverband:coverage` and see no output to reproduce.

thx.

kbaum:
Just thinking about this one. Would it make sense to support this task by just starting a rack server locally and serving the report in the same way we do when it's embedded in rails app? This might simplify things.

danmayer:
hmm interesting idea @kbaum could be nice in terms of having a single experience. 

honestly, the fix is easy I am more concerned with figuring out some good tests to catch this... The tests around the reporting need to be reworked as we add more front end complexity.

The fix is to pass in a new var from the HTML Reporter, to set if we are static or dynamic... `static = true` or something.

Then update https://github.com/danmayer/coverband/blob/master/views/layout.erb#L51 to something like:

```
  <%= static ? formatted_source_file(result, source_file) : formatted_source_file_loader(result, source_file) %>
```

danmayer:
I guess not well named, but we already from the reporter call two different paths and we can tell static or dynamic based on `base_path` this is already used to handle how we render assets differently

https://github.com/danmayer/coverband/blob/master/lib/coverband/utils/html_formatter.rb#L79

kbaum:
If it's a quick fix, i can see just moving forward with it for now. In terms of testing, we could punt and longer term move towards reporting always being a rack server. Then we could go with a capybara style feature test for it. 

The downside of having only a rack server is that we wouldn't be supporting use cases of serving the report from s3 or a continuous build services.  In my case since our rails instance in production is api only, I would just run the rack server locally and point REDIS_URL at production. That works assuming REDIS is not behind a production firewall.

danmayer:
resolved

PR #245, opened 2019-05-02
Fix static html

danmayer:
* check if we are generated dynamic or static html
* remove the try usage, which was Rails only, oops ;(
* add better test coverage around our HTML generation as it is getting more complicated
* handle file has been removed issue

This addresses issue #234


lib/coverband/utils/source_file.rb
```diff
@@ -175,6 +179,14 @@ def covered_lines
         @covered_lines ||= lines.select(&:covered?)
       end
 
+      def covered_lines_count
+        covered_lines ? covered_lines.count : nil
+      end
+
+      def line_coverage(index)
+        lines[index] ? lines[index].coverage : nil
```
kbaum:
```ruby
line[index]&.coverage
```

lib/coverband/utils/source_file.rb
```diff
@@ -175,6 +179,14 @@ def covered_lines
         @covered_lines ||= lines.select(&:covered?)
       end
 
+      def covered_lines_count
+        covered_lines ? covered_lines.count : nil
```
kbaum: Instead of the ternary?
```ruby
covered_lines&.count
```

test/coverband/utils/html_formatter_test.rb
```diff
@@ -0,0 +1,45 @@
...
+    Coverband.configure do |config|
+      config.store             = @store
+      config.s3_bucket         = nil
+      config.ignore            = ['notsomething.rb']
+    end
+    mock_file_hash
+    @store.send(:save_report, basic_coverage_full_path)
+
+    filtered_report_files = Coverband::Reporters::Base.report(@store, {})
+    Coverband::Utils::HTMLFormatter.new(filtered_report_files).format!
```
kbaum: One thought and not something we need to fix in this PR. Our api does not make it obvious when we are generating a static html report vs when we are hosting the report.

lib/coverband/utils/source_file.rb
```diff
@@ -152,6 +152,10 @@ def covered_percent
         Float(covered_lines.size * 100.0 / relevant_lines.to_f)
       end
 
+      def formatted_covered_percent
+        covered_percent ? covered_percent.round(2) : nil
```
kbaum:
```ruby
covered_percent&.round(2)
```

lib/coverband/utils/source_file.rb
danmayer: good call fixed all of these

test/coverband/utils/html_formatter_test.rb
danmayer: yeah agreed, I think there are a couple refactorings to do on HTMLReport and HTMLFormatter I have sprinkled some `todos` about a few cleanups and I think making it more clear what we are generating would be good

PR #246, opened 2019-05-03
fix bug which caused files with 0 load or runtime coverage to not sho…

### LINKED ISSUE #242: Coverband::Reporters::HTMLReport.fix_reports method is incorrectly setting filtered_report_files
The [`fix_reports`](https://github.com/danmayer/coverband/blob/v4.2.0/lib/coverband/reporters/html_report.rb#L51-L61) method is setting the value of `self.filtered_report_files` to `reports` instead of `filtered_report_files` so it is missing all the files that were not tracked by coverband. This causes the generated report to be inaccurate and only count files that were tracked by coverband.

The `fix_reports` method should look like this:

```
def self.fix_reports(reports)
  # list all files, even if not tracked by Coverband (0% coverage)
  tracked_glob = "#{Coverband.configuration.current_root}/{app,lib,config}/**/*.{rb}"
  filtered_report_files = {}

  reports.each_pair do |report_name, report_data|
    filtered_report_files[report_name] = {}
    report_files = Coverband::Utils::Result.add_not_loaded_files(report_data, tracked_glob)

    # apply coverband filters
    report_files.each_pair do |file, data|
      next if Coverband.configuration.ignore.any? { |i| file.match(i) }

      filtered_report_files[report_name][file] = data
    end
  end
  filtered_report_files
end
```

danmayer:
…w in reports

Thanks @ckumar1  this resolves #242 the update you mentioned is correct, this just adds a test to ensure there isn't another regression.


README.md
```diff
@@ -335,6 +335,7 @@ If you submit a change please make sure the tests and benchmarks are passing.
 - view test coverage: `open coverage/index.html`
 - run the benchmarks before and after your change to see impact
   - `rake benchmarks`
+  - run a single test by line number like rspec: `bundle exec m test/coverband/reporters/html_test.rb:29`
```
kbaum: Great idea!

PR #250, opened 2019-05-05
fix #249 where clear and coverage add coverage

### LINKED ISSUE #249: Clearing coverage generates coverage
Start with an empty redis using redis-cli:

```
redis-cli
127.0.0.1:6379> flushdb
OK
```
Execute clear coverage task:
```
bundle exec rake coverband:clear
```
Then look in redis using cli:
```
redis-cli
127.0.0.1:6379> get "coverband_3_2"
```
And there will be some data.

The same issue occurs with `coverband:coverage`.

This can be very problematic if someone is generating coverage from their local machine while pointing at production. In my case this is needed since our rails server is api only and I cannot host the report within the production application. I end up writing coverage data from my local dev machine into production.

danmayer:
OK so this should fix the issue you pointed out @kbaum but I think it highlighted another issue take a look at the comment I put here https://github.com/danmayer/coverband/compare/fix_rake_at_exit_issue?expand=1#diff-245138fa5d6754b051eb62e3472bd791R4

and let me know if you agree / understand what I am bringing up... if so I can work on resolving that in another PR as well.

In terms of testing, it is kind of hard to test the Rake tasks, I think this gets back to we need an "integration" test suite that runs in different processes and verifies outputs.

### DIFF
--- a/lib/coverband/at_exit.rb
+++ b/lib/coverband/at_exit.rb
@@ -24,7 +24,10 @@ module Coverband
           # * Perhaps detect heroku deployment ENV var opposed to tasks?
           #####
           default_heroku_tasks = ['assets:clean', 'assets:precompile']
-          if defined?(Rake) && Rake.respond_to?(:application) && (Rake&.application&.top_level_tasks || []).any? { |task| default_heroku_tasks.include?(task) }
+          if !Coverband.configuration.report_on_exit ||
+             (defined?(Rake) &&
+             Rake.respond_to?(:application) &&
+             (Rake&.application&.top_level_tasks || []).any? { |task| default_heroku_tasks.include?(task) })
             # skip reporting
           else
             Coverband.report_coverage(true)
--- a/lib/coverband/utils/tasks.rb
+++ b/lib/coverband/utils/tasks.rb
@@ -11,6 +11,7 @@ namespace :coverband do
 
   def environment
     Coverband.configuration.report_on_exit = false
+    Coverband.configuration.background_reporting_enabled = false
     Rake.application['environment'].invoke if Rake::Task.task_defined?('environment')
   end
 
--- a/lib/coverband/utils/tasks.rb
+++ b/lib/coverband/utils/tasks.rb
@@ -18,7 +18,7 @@ namespace :coverband do
   task :coverage do
     environment
     if Coverband.configuration.reporter == 'scov'
-      Coverband::Reporters::HTMLReport.report(Coverband.configuration.store)
+      Coverband::Reporters::HTMLReport.new(Coverband.configuration.store).report
     else
       Coverband::Reporters::ConsoleReport.report(Coverband.configuration.store)
     end
--- a/lib/coverband/utils/tasks.rb
+++ b/lib/coverband/utils/tasks.rb
@@ -1,9 +1,16 @@
 # frozen_string_literal: true
 
 namespace :coverband do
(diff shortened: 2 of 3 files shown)


kbaum:
Will take a look shortly. Was spiking on forking some tests this morning:

https://github.com/danmayer/coverband/compare/fork_integration_tests

The idea would be to fork only specific tests so we still are able to report test coverage.

lib/coverband/utils/tasks.rb
```diff
@@ -1,17 +1,24 @@
...
+  # TODO: FixMe before releasing 4.2.1
+  # This was previously needed when Coverband was generally require false
+  # now this would double configure, which isn't a good thing
+  # I think we need ot make it safe to call configure twice and no-op if it has been called
+  # because removing this means if you have require: false on coverband or
+  # COVERBAND_DISABLE_AUTO_START=true the rake tasks wouldn't work
+  # Coverband.configure
 
   def environment
+    Coverband.configuration.report_on_exit = false
```
kbaum: Think we should disable background thread also.

lib/coverband/at_exit.rb
```diff
@@ -24,7 +24,10 @@ def self.register
           # * Perhaps detect heroku deployment ENV var opposed to tasks?
           #####
           default_heroku_tasks = ['assets:clean', 'assets:precompile']
-          if defined?(Rake) && Rake.respond_to?(:application) && (Rake&.application&.top_level_tasks || []).any? { |task| default_heroku_tasks.include?(task) }
+          if !Coverband.configuration.report_on_exit ||
+             (defined?(Rake) &&
+             Rake.respond_to?(:application) &&
+             (Rake&.application&.top_level_tasks || []).any? { |task| default_heroku_tasks.include?(task) })
```
kbaum: Since we are setting the flag within the environment task, are these rake tasks still necessary? I believe all of the heroku tasks depend on the environment rake task.

lib/coverband/utils/tasks.rb
danmayer: good call done

lib/coverband/at_exit.rb
danmayer: so this isn't the default environment just a method called `environment` in the Coverband rake task. I think we could hook into the rake tasks environment used by things on Heroku but if we did the issue would be how can we report on rake tasks used as part of cron jobs, but not report on preloading tasks like `assets`. The issue this prevented was on deploy having the asset job ending up putting a bunch of data in runtime not load. We could likely revisit as we learned rake skips eager loading but I think that is still different than our internal rake tasks which we wan to do nothing at all.

kbaum: Great point about cron job rake tasks. I think we should leave this for now. That being said, it's tough to remember what specific data was being added to `runtime` coverage and why that was a problem. Now that we have forked tests (#251), we should probably look to add a test that exposes the need for this code. For another PR though.

danmayer: yeah agreed we need to definitely simplify and some of this underlying logic has changed I added a comment on this PR as it was a pretty fundamental change to how all this stuff works.

https://github.com/danmayer/coverband/pull/238

 Perhaps opposed to hard coded Heroku rake tasks we could move to a config `config.rake_tasks_to_ignore` similar to how we treat ignoring files, and it could have some nice defaults like the heroku rake tasks.

PR #251, opened 2019-05-05
Fork rails full stack test

kbaum:
Any test placed within the `test/forked` directory are forked for each example.

Add a `forked_tests` rake task for running forked tests separately from the other tests.


kbaum:
Great thank!

PR #253, opened 2019-05-08
Changed middleware to not report

kbaum:
Since we have moved to a background thread for reporting, this removes reporting from middleware. I think this will help to avoid confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ For the best coverage you want this loaded as early as possible. I have been put
 require 'coverband'
 require File.dirname(__FILE__) + '/config/environment'
 
-use Coverband::Middleware
+use Coverband::BackgroundMiddleware
 run ActionController::Dispatcher.new
 ```
 
--- a/lib/coverband.rb
+++ b/lib/coverband.rb
@@ -26,7 +26,7 @@ require 'coverband/reporters/console_report'
 require 'coverband/integrations/background'
 require 'coverband/integrations/rack_server_check'
 require 'coverband/reporters/web'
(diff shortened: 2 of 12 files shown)


danmayer:
I have some things I would still like to work on that require middleware, I haven't gotten to them given the current loading related feature set. So I would prefer if we are going to do this to split it to the default middleware doesn't report but there is a report middleware one can opt into, that way I don't have to reimplement it and the tests when adding the functionality.

In the recent example, I don't think there was confusion over the current middleware, but there was custom middleware involved that overrode it.

danmayer:
Yeah, definitely like reducing branching and scope, I can perhaps pull this into the jam session gem I have been talking about to keep experiments out of the mainline. As we pulled out SimpleCov the complexity in the gem has increased, it has opened up a lot of features, but does mean we need to spend time keeping things as simple as we can. I appreciate you always pushing that way. 

I am also going to try to push to reduce complexity to take full advantage of all new building blocks we have to avoid the complexity that we no longer should need to consider. 

This looks good to merge, I can extract out the new class a bit later

kbaum:
Ok great. I might have misunderstood, but i think I've already extracted out the new class in ReportMiddleware:

https://github.com/danmayer/coverband/pull/253/files#diff-599f49677e635a02d60dded7c2578237

> I am also going to try to push to reduce complexity to take full advantage of all new building blocks we have to avoid the complexity that we no longer should need to consider.

It's kind of meta but I think this fits with the reasoning behind coverband itself 😄

kbaum:
Ahh.. gotcha. Makes sense.

PR #257, opened 2019-05-08
Remove force_report

kbaum:
Removed force_coverage logic as I think we can always report coverage since we are using background reporting.

### DIFF
--- a/lib/coverband/collectors/coverage.rb
+++ b/lib/coverband/collectors/coverage.rb
@@ -43,9 +43,9 @@ module Coverband
         end
       rescue StandardError => e
         if @verbose
-          @logger&.error 'coverage failed to store'
-          @logger&.error "error: #{e.inspect} #{e.message}"
-          @logger&.error e.backtrace
+          @logger.error 'coverage failed to store'
+          @logger.error "error: #{e.inspect} #{e.message}"
+          @logger.error e.backtrace
         end
         raise e if @test_env
       end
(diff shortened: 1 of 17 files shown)


lib/coverband/collectors/coverage.rb
```diff
@@ -77,10 +75,6 @@ def filtered_files(new_results)
         end.select { |_file_name, coverage| coverage.any? { |value| value&.nonzero? } }
       end
 
-      def ready_to_report?
```
danmayer: I thought we still allowed this for resque polling? Not opposed to removing though

danmayer: if we remove this I think we can remove @reporting_frequency and the coverband config related to it

danmayer:
Nice the logger issue has come up here and there, I wonder if in our new railtie setup if we are initializing before `Rails.logger` has initialized? Perhaps we have to more lazily load it?

lib/coverband/collectors/coverage.rb
kbaum: Double checked and I already deleted from configuration.

> I thought we still allowed this for resque polling?

Doesn't look like we are using this within our resque integration as we were passing true for force_report. Was thinking about it and I feel like this configuration is confusing when used along with background reporting which reports every 30 seconds by default and has it's own `Configuration#background_reporting_sleep_seconds`. Perhaps we remove it for now and bring something like it back in another form if needed?

danmayer: yeah agreed and it is really an edge case and if we want to simplify we just have to make calls, on a pretty high traffic resque installation we have been fine at 100% coverage, it does slow down the background jobs but not much can be done about that with the forking model and the coverage was worth it for us.

kbaum: Ok sounds good.

PR #258, opened 2019-05-09
Add upgrade section

kbaum:


kbaum:
Looks like we have one sporadically failing test:

image

PR #264, opened 2019-05-17
Oneshot coverage

zwalker:
Adds a configuration option `use_oneshot_lines_coverage` which when true enables oneshot_lines coverage in ruby versions >= 2.6.

### DIFF
--- a/lib/coverband/collectors/delta.rb
+++ b/lib/coverband/collectors/delta.rb
@@ -12,7 +12,17 @@ module Coverband
 
       class RubyCoverage
         def self.results
-          ::Coverage.peek_result.dup
+          results = ::Coverage.peek_result.dup
+          Coverband.configuration.use_oneshot_lines_coverage ? transform_oneshot_lines_results(results) : results
+        end
+
+        def self.transform_oneshot_lines_results(results)
+          results.each_with_object({}) do |(file, coverage), new_results|
+            line_counts = coverage[:oneshot_lines].each_with_object(::Coverage.line_stub(file)) do |line_number, line_counts|
+              line_counts[line_number - 1] = 1
+            end
+            new_results[file] = line_counts
+          end
         end
       end
 
--- a/lib/coverband/collectors/delta.rb
+++ b/lib/coverband/collectors/delta.rb
@@ -12,22 +12,14 @@ module Coverband
 
       class RubyCoverage
         def self.results
-          results = ::Coverage.peek_result.dup
-          Coverband.configuration.use_oneshot_lines_coverage ? transform_oneshot_lines_results(results) : results
-        end
-
-        def self.transform_oneshot_lines_results(results)
-          results.each_with_object({}) do |(file, coverage), new_results|
-            line_counts = coverage[:oneshot_lines].each_with_object(::Coverage.line_stub(file)) do |line_number, line_counts|
-              line_counts[line_number - 1] = 1
-            end
-            new_results[file] = line_counts
-          end
+          ::Coverage.peek_result.dup
         end
       end
 
       def self.results(process_coverage = RubyCoverage)
-        new(process_coverage.results).results
+        coverage_results = process_coverage.results
+        coverage_results = transform_oneshot_lines_results(coverage_results) if Coverband.configuration.use_oneshot_lines_coverage
+        new(coverage_results).results
       end
 
       def results
@@ -55,6 +47,15 @@ module Coverband
       def array_diff(latest, original)
         latest.map.with_index do |v, i|
           [0, v - original[i]].max if v && original[i]
+        end
+      end
+
+      def self.transform_oneshot_lines_results(results)
+        results.each_with_object({}) do |(file, coverage), new_results|
+          line_counts = coverage[:oneshot_lines].each_with_object(::Coverage.line_stub(file)) do |line_number, line_counts|
+            line_counts[line_number - 1] = 1
+          end
+          new_results[file] = line_counts
         end
       end
     end
--- a/lib/coverband/collectors/delta.rb
+++ b/lib/coverband/collectors/delta.rb
@@ -50,12 +50,12 @@ module Coverband
         end
       end
 
-      def self.transform_oneshot_lines_results(results)
+      private_class_method def self.transform_oneshot_lines_results(results)
         results.each_with_object({}) do |(file, coverage), new_results|
-          line_counts = coverage[:oneshot_lines].each_with_object(::Coverage.line_stub(file)) do |line_number, line_counts|
+          transformed_line_counts = coverage[:oneshot_lines].each_with_object(::Coverage.line_stub(file)) do |line_number, line_counts|
             line_counts[line_number - 1] = 1
           end
-          new_results[file] = line_counts
+          new_results[file] = transformed_line_counts
         end
       end
     end
--- a/lib/coverband/configuration.rb
+++ b/lib/coverband/configuration.rb
@@ -154,7 +154,8 @@ module Coverband
     end
 
     def use_oneshot_lines_coverage=(value)
-      raise Exception.new('One shot line coverage is only available in ruby >= 2.6') unless one_shot_coverage_implemented_in_ruby_version? || !value
+      raise(Exception, 'One shot line coverage is only available in ruby >= 2.6') unless one_shot_coverage_implemented_in_ruby_version? || !value
+
       @use_oneshot_lines_coverage = value
     end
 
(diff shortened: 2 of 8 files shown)


lib/coverband/collectors/delta.rb
```diff
@@ -12,7 +12,17 @@ def initialize(current_coverage)
 
       class RubyCoverage
         def self.results
-          ::Coverage.peek_result.dup
+          results = ::Coverage.peek_result.dup
+          Coverband.configuration.use_oneshot_lines_coverage ? transform_oneshot_lines_results(results) : results
+        end
+
+        def self.transform_oneshot_lines_results(results)
```
kbaum: I think want to keep our logic outside of this class. It's purpose is to allow for stubbing out ruby coverage for testing using dependency injection.

https://github.com/danmayer/coverband/blob/master/test/coverband/collectors/delta_test.rb

test/coverband/collectors/delta_test.rb
```diff
@@ -48,4 +48,32 @@ def setup
     results = Coverband::Collectors::Delta.results(mock_coverage(current_coverage))
     assert_equal(current_coverage, results)
   end
+
+  test 'one shot lines results' do
```
kbaum: Assuming we change [transform_oneshot_line_coverage](https://github.com/danmayer/coverband/pull/264/files#r285338414) somewhere else, I think we can test this without mocking following the same approach as the other tests in this file. Let me know what you think.

test/coverband/configuration_test.rb
```diff
@@ -86,4 +86,23 @@ def setup
       end
     end
   end
+
+  test 'use_oneshot_lines_coverage' do
+    Coverband::Collectors::Coverage.instance.reset_instance
```
kbaum: Do we need to reset here? Was under impress that our setup does this (I hope).

test/coverband/configuration_test.rb
```diff
@@ -86,4 +86,23 @@ def setup
...
+    refute Coverband.configuration.use_oneshot_lines_coverage
+
+    Coverband.configuration.stubs(:one_shot_coverage_implemented_in_ruby_version?).returns(true)
```
kbaum: Can we avoid stubbing here? I think it would be possible by doing each of these assertions within a separate example/test_method.

kbaum: Not for this PR, but I think we should try to avoid stubbing in the class under test. I think with some refactoring, we could stick all of the ruby version checks in another class and inject our own class double within our tests.

danmayer:
Nice, this has been on the list for a little while... The biggest concern I have is trying to figure out how we deal with this and the new initialize vs runtime functionality. 

Thoughts if these modes would conflict @kbaum & @zwalker 

I am not sure they are entirely compatible:
* Anything hit in initialization vs one shot would never get hit during runtime
    * This might be confusing
    * we could disable all the runtime features when the oneshot mode is on...
* Perhaps it isn't really an issue as method definitions and such get hit at initialization but method bodies are hit at runtime and means this doesn't really overlap

lib/coverband/collectors/delta.rb
danmayer: yeah probably best to just put it in the primary Coverage collector class

danmayer:
This is a great start, I ran this locally and it showed me that I think it plays fine with the current plans for the initialize vs eager loading. So I think with a bit of cleanup we can get this in... Do you want to try to refactor stuff out of the `Delta` class or want me to give it a shot @zwalker

zwalker:
Thanks for the comments.  I'll give a go at making the requested changes.

lib/coverband/collectors/delta.rb
zwalker: @danmayer Can you point me to the class you are referring to here as the "primary Coverage collector class".  Do you mean Coverband::Collectors::Delta?

kbaum: Think he means [coverage.rb](https://github.com/danmayer/coverband/blob/master/lib/coverband/collectors/coverage.rb)

zwalker: I moved the into the Delta class as that was my best guess at what you were suggesting

test/coverband/collectors/delta_test.rb
zwalker: 👍

test/coverband/configuration_test.rb
zwalker: Removed

lib/coverband/collectors/delta.rb
zwalker: @kbam, I didn't see your response until after I had pushed a commit.
I think the transform_oneshot_lines_results method needs to be used in the Delta class because it is comparing to the previous result which is also stored in the Delta class, so I left it there.

I can move it to Coverband::Collectors::Coverage if you that is the preference, but I think the usage would still stay in Delta.

test/coverband/collectors/coverage_test.rb
```diff
@@ -70,4 +70,38 @@ def teardown
...
+    Coverband::Collectors::Coverage.send(:new)
+  end
+
+  test 'no one shot line coverage for ruby < 2.6' do
+    return unless Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.5.0')
+    Coverband::Collectors::Coverage.expects(:ruby_version_greater_than_or_equal_to?).with('2.6.0').returns(false)
+    Coverband::Collectors::Coverage.expects(:ruby_version_greater_than_or_equal_to?).with('2.5.0').returns(true)
+    ::Coverage.expects(:running?).returns(false)
+    ::Coverage.expects(:start).with()
+    Coverband::Collectors::Coverage.send(:new)
```
kbaum: Instead of invoking a private method, could we just calling `Coverage.reset_instance`?

test/coverband/collectors/delta_test.rb
```diff
@@ -48,4 +48,30 @@ def setup
...
+    unless Coverband.configuration.one_shot_coverage_implemented_in_ruby_version?
+      module ::Coverage
```
kbaum: Instead of monkey patching here, maybe just not run this test in ruby < 2.6 ?

lib/coverband/configuration.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'rubygems'
```
kbaum: Don't think this is needed.

test/coverband/collectors/coverage_test.rb
zwalker: `reset_instance` doesn't seem to actually call `initialize` for a new instance so in this case, I don't think exercises the code this test wants to exercise.

test/coverband/collectors/delta_test.rb
zwalker: It's not clear to me where the setup is for what versions of ruby are tested.  As long as it is actually tested on versions of ruby >= 2.6, that seems fine.

What did you have in mind though?
something like
```
if Coverband.configuration.one_shot_coverage_implemented_in_ruby_version?
  test 'one shot lines results' do
    ...
  end
end
```

lib/coverband/configuration.rb
zwalker: I think it is required for old versions of Ruby, like 1.9 maybe.  If we aren't worried about going back that far then it can be removed.

kbaum: Right. Coverband won’t work that far back anyway.

test/coverband/collectors/delta_test.rb
kbaum: Yes, I think that is simpler.

lib/coverband/collectors/delta.rb
danmayer: yeah I was thinking `collectors/coverage` let me give this PR another close look tomorrow. I should have a bit more time to dig in then

lib/coverband/configuration.rb
zwalker: Done

test/coverband/collectors/delta_test.rb
zwalker: Done, the test is now run conditionally.

danmayer:
OK merging this, thanks so much @zwalker we will get this into the next release

kbaum:
This is great! We should add to readme and perhaps get some meaningful benchmarks. Thanks @zwalker!

zwalker:
Great.  Thank you!

Issue #261, opened 2019-05-15
Runtime vs Eager not working for Gems in Rails application


kbaum:
When tracking gems in production, I see that overall coverage and runtime coverage are the same:

![image](https://user-images.githubusercontent.com/96786/57771281-44c4a280-76e0-11e9-9ef8-8d13f10187d1.png)

Looking at the details, I can see that loading the class definitions is marked as runtime:

![image](https://user-images.githubusercontent.com/96786/57771417-a97ffd00-76e0-11e9-8f4f-e6e8d0fd5861.png)

My guess is that bundler does not load gems as part of rails eager_load. Since our [current implementation](https://github.com/danmayer/coverband/blob/master/lib/coverband/utils/railtie.rb#L6) relies on rails eager_load to track eager_load coverage, gems are left out of this functionality.

Not sure of the right way to handle this, but being able to track gem usage at runtime would be a great feature as I think removing gems can dramatically reduce a rails process memory footprint and startup time.

We could attempt to capture bundle require as part of eager_load data? This might be confusing though as it is not technically part of rails eager_loading. This makes me think that we got the name wrong and perhaps what we want to capture is "initialization" coverage data vs "runtime".

kbaum:
So I don't think my guess is correct. I was having difficulties reproducing locally even though this was happening in our heroku production environment. I cleared redis and set track_gems to true in production and I still saw the issue. I then cleared coverage again in production and restart one more time and the issue is no longer happening. Can't explain what I am seeing and it is difficult to trace down since I cannot reproduce consistently on my local machine.

danmayer:
so another thing on this one was I know after the ignore first report I didn't see this so we might still be missing something on handling 'initialize' vs runtime that is just a bigger issue than eager_loading (as you have mentioned)

kbaum:
So looks like this problem happens when using a rake task that depends on environment. This makes sense given what we've learned previously. Eager loading is disabled but application.rb requires all of the Gems:

```ruby
Bundler.require(*Rails.groups)
```

This is essentially an eager load of all of those files on initialization. To address, I added this within application.rb:

```ruby
require 'coverband'
Coverband.eager_loading_coverage!
Bundler.require(*Rails.groups)
Coverband.report_coverage
Coverband.runtime_coverage!
```

Thinking about how we would address this issue for everyone else and might be difficult given `Bundle.require(*Rails.groups)` happens before coverband is required. This may require users to modify their application.rb.

kbaum:
[...]

PR #265, opened 2019-05-20
Add Bundle.require eager_loading

kbaum:
Modifies rails apps to invoke eager loading within an `eager_loading` block.

Also adds a test that ensures gem coverage is handled correctly.

### DIFF
--- a/Gemfile.rails4
+++ b/Gemfile.rails4
@@ -7,3 +7,4 @@ gemspec
 gem 'rails', '~>4'
 # this is used for testing gem tracking
 gem 'rainbow', require: false
+gem 'pundit'
--- a/lib/coverband.rb
+++ b/lib/coverband.rb
@@ -93,14 +93,7 @@ module Coverband
     configure
     start
     require 'coverband/utils/railtie' if defined? ::Rails::Railtie
-    require 'coverband/integrations/resque' if defined? Resque
-    if defined?(Bundler)
-      module BundlerEagerLoad
-        def require(*groups)
-          Coverband.eager_loading_coverage { super }
-        end
-      end
-      Bundler.singleton_class.prepend BundlerEagerLoad
-    end
+    require 'coverband/integrations/resque' if defined? ::Resque
+    require 'coverband/integrations/bundler' if defined? ::Bundler
   end
 end
--- a/lib/coverband.rb
+++ b/lib/coverband.rb
@@ -76,6 +76,10 @@ module Coverband
     coverage_instance.eager_loading!
   end
 
+  def self.eager_loading_coverage(&block)
+    coverage_instance.eager_loading(&block)
+  end
+
   def self.runtime_coverage!
     coverage_instance.runtime!
   end
--- a/lib/coverband.rb
+++ b/lib/coverband.rb
@@ -94,5 +94,13 @@ module Coverband
     start
     require 'coverband/utils/railtie' if defined? ::Rails::Railtie
     require 'coverband/integrations/resque' if defined? Resque
+    if defined?(Bundler)
+      module BundlerEagerLoad
+        def require(*groups)
+          Coverband.eager_loading_coverage { super }
+        end
+      end
+      Bundler.singleton_class.prepend BundlerEagerLoad
+    end
   end
 end
--- a/lib/coverband/collectors/coverage.rb
+++ b/lib/coverband/collectors/coverage.rb
@@ -38,6 +38,15 @@ module Coverband
         @store.type = Coverband::EAGER_TYPE
       end
 
+      def eager_loading(&block)
+        old_coverage_type = @store.type
+        eager_loading!
+        block.call
+      ensure
+        report_coverage
+        @store.type = old_coverage_type
+      end
+
       def report_coverage
         @semaphore.synchronize do
           raise 'no Coverband store set' unless @store
--- a/test/forked/rails_rake_full_stack_test.rb
+++ b/test/forked/rails_rake_full_stack_test.rb
@@ -1,0 +1,22 @@
+require File.expand_path('../rails_test_helper', File.dirname(__FILE__))
+require 'rails'
+
+class RailsRakeFullStackTest < Minitest::Test
+
+  test 'rake tasks shows coverage properly within eager_loading' do
+    system("COVERBAND_CONFIG=./test/rails#{Rails::VERSION::MAJOR}_dummy/config/coverband.rb bundle exec rake -f test/rails#{Rails::VERSION::MAJOR}_dummy/Rakefile middleware")
+    store.type = :eager_loading
+    pundit_file = store.coverage.keys.grep(/pundit.rb/).first
+    refute_nil pundit_file
+    pundit_coverage = store.coverage[pundit_file]
+    refute_nil pundit_coverage
+    assert_includes pundit_coverage['data'], 1
+
+    store.type = nil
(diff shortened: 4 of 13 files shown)


Gemfile.rails4
```diff
@@ -7,3 +7,4 @@ gemspec
 gem 'rails', '~>4'
 # this is used for testing gem tracking
 gem 'rainbow', require: false
+gem 'pundit'
```
kbaum: Needed to add another gem since rainbow is require false here. Not sure if it needs to be require: false?

lib/coverband/collectors/coverage.rb
```diff
@@ -34,6 +34,15 @@ def eager_loading!
         @store.type = Coverband::EAGER_TYPE
       end
 
+      def eager_loading(&block)
```
kbaum: Still need a test for this method but wanted to get the idea out there.

test/forked/rails_rake_full_stack_test.rb
```diff
@@ -0,0 +1,21 @@
+require File.expand_path('../rails_test_helper', File.dirname(__FILE__))
+
+class RailsRakeFullStackTest < Minitest::Test
+
+  test 'rake tasks shows coverage properly within eager_loading' do
+    system("COVERBAND_CONFIG=./test/rails5_dummy/config/coverband.rb bundle exec rake -f test/rails5_dummy/Rakefile middleware")
```
kbaum: Shelling out allows us to test rails initialization from a rake task.

test/rails4_dummy/config/application.rb
```diff
@@ -4,14 +4,11 @@
 
 require 'rails'
 require 'action_controller/railtie'
+require 'coverband'
+Coverband.eager_loading_coverage { Bundler.require(*Rails.groups) }
```
kbaum: We could remove the need for this line by patching Bundler.require. I think we would still need an early require of coverband within rails application.rb though since coverband has to be loaded before `Bundler.require` is called.

Gemfile.rails4
danmayer: I think I did that on purpose on rainbow but we could likely revisit it as a lot has changed out how we load initial Coverage since then

test/rails4_dummy/config/application.rb
danmayer: hmm so with this method we would as part of installation require folks to modify the application.rb...

Kinda of hurts the zero config railties story...

So the issue we are trying to solve is https://github.com/danmayer/coverband/issues/261

Thinking a bit more of other ways to address still feels like we just aren't covering all cases related to initialization... 

Kind of feels like each solution has had a slightly different bug story... We started with 

`config.after_initialize` and `before_initialize` which I think would work for the bundler require command... is some combination of eager_load, and the initializer hooks something that could cover all scenarios... A case were more detailed bug reports on each of the issues would have been good.... Rake was disabling eager loading... the prepend solved that... After initialize was adding code we didn't want to load time vs runtime... That was avoided by ignoring first request... 

Anyways, just feel like we are still missing how to be cover all cases. I guess this PR works, but I don't love the solution.

kbaum: > Kinda of hurts the zero config railties story...

Agreed. I'm not happy about that either. I'm just not sure how we cleanly capture `Bundle.require` without inserting some code before the `Bundle.require`. The problem is that the `Bundle.require` itself is what requires the coverband library. We are tracking the code that requires the tracking code.

> Kind of feels like each solution has had a slightly different bug story...

Agreed that we are kind of playing whack a mole with each solution. I think that it's important that going forward we put the extra effort in to have a test case to reproduce each issue. Then we will evolve toward the best solution that makes all the tests pass.

> I guess this PR works, but I don't love the solution.

This PR is really only an example of how one could track runtime/eager_loading coverage of gems. It also doesn't really change the coverband library itself other than adding the [Coverband#eager_loading](https://github.com/danmayer/coverband/pull/265/files#diff-181f764c69748a3ceb0cc64e4a37ebaeR37) method taking a block which I kind of like. Maybe we can evolve to a place where we don't need custom code for tracking runtime/eager_loading coverage of gems but I think this takes us in a positive direction.

kbaum: Added a [commit](https://github.com/danmayer/coverband/pull/265/commits/4bca5e51ffc981bc66bc626b380c5fb64c5b9669) that moves most custom code out of application. Now all that's left for tracking `Bundle#require` of gems accurately is requiring coverband before the call to `Bundle#require`. This makes sense to me. Another thought is that coverband still works out of box with rails, but if you would like to track the require of gems, coverband must be loaded before hand.

lib/coverband.rb
```diff
@@ -90,5 +94,13 @@ def self.runtime_coverage!
     start
     require 'coverband/utils/railtie' if defined? ::Rails::Railtie
     require 'coverband/integrations/resque' if defined? Resque
+    if defined?(Bundler)
```
danmayer: perhaps move into `coverband/integrations/bundler` with the same pattern above?

test/forked/rails_rake_full_stack_test.rb
danmayer: This is awesome, I think we can add a few more like this (not in this PR) to cover all the weird load order bugs we have been sorting out.

lib/coverband.rb
kbaum: Agreed.

test/rails4_dummy/config/application.rb
danmayer: > Agreed that we are kind of playing whack a mole with each solution. I think that it's important that going forward we put the extra effort in to have a test case to reproduce each issue. Then we will evolve toward the best solution that makes all the tests pass.

yep agreed, some of these are hard to reproduce in tests, but I think with the new forked tests model we should be able to figure out a way.

danmayer:
OK, I added one comment on a small refactoring...

Then I guess update the docs to make it clear to manually require coverband if using gem tracking. I think since I am still considering Gem tracking in 'alpha' release (off by default). It is OK. I am hoping we can perhaps come up with something even more clever that wouldn't require this... I am wondering for example if we required coverband to be loaded at the top of the gemfile we could capture this.

one problem with gem tracking in general if we need coverband to be required AFTER resque for that integration to work... So there are some kinks to workout still.

test/forked/rails_rake_full_stack_test.rb
kbaum: Agreed. Whenever we think of tests we are missing, let's add them to issues

lib/coverband.rb
kbaum: Done.

kbaum:
Makes sense that gem tracking of eager/runtime is tricky to pull off without a little bit of manual code. Let me know if it's ok to merge this one. Thx.

danmayer:
OK merging this @kbaum but could you update the readme with how to ensure Gems works with eager loading, you can just commit that to master

kbaum:
Will do 👍

kbaum:
@danmayer Added instructions for requiring coverband within application.rb for gem tracking:

https://github.com/danmayer/coverband/commit/5725e7471ef72c8aab7ce25b1420a0873aeb41c6

We don't yet reference eager/runtime loading within the README so I'll need to come back to that. I'll create an issue for that.

PR #270, opened 2019-05-25
Spike on runtime relevant lines

danmayer:
take a look at this spike and let me know your thoughts @kbaum it definitely adds some complexity, but I can't think of a much better way to allow a source file to know data about a source file from a different execution time context.

This needs further cleanup but wanted to see if the approach makes sense.

### DIFF
--- a/lib/coverband/utils/results.rb
+++ b/lib/coverband/utils/results.rb
@@ -30,7 +30,7 @@ module Coverband
 
         return 0.0 unless runtime_file
 
-        return 100.0 unless eager_file
+        return runtime_file.formatted_covered_percent unless eager_file
 
         runtime_relavant_lines = eager_file.relevant_lines - eager_file.covered_lines_count
         runtime_file.runtime_relavant_calculations(runtime_relavant_lines) { |file| file.formatted_covered_percent }
--- a/lib/coverband/utils/results.rb
+++ b/lib/coverband/utils/results.rb
@@ -20,6 +20,22 @@ module Coverband
         get_results(results_type).source_files.find { |file| file.filename == source_file.filename }
       end
 
+      def runtime_relevant_coverage(source_file)
+        eager_loading_coverage = get_results(Coverband::EAGER_TYPE)
+        runtime_coverage = get_results(Coverband::RUNTIME_TYPE)
+        return unless eager_loading_coverage && runtime_coverage
+
+        eager_file = eager_loading_coverage.source_files.find { |file| file.filename == source_file.filename }
+        runtime_file = runtime_coverage.source_files.find { |file| file.filename == source_file.filename }
+
+        return 0.0 unless runtime_file
+
+        return 100.0 unless eager_file
+
+        runtime_relavant_lines = eager_file.relevant_lines - eager_file.covered_lines_count
+        runtime_file.runtime_relavant_calculations(runtime_relavant_lines) { |file| file.formatted_covered_percent }
+      end
+
       ###
       # TODO: Groups still have some issues, this should be generic for groups, but right now gem_name
       # is specifically called out, need to revisit all gorups code.
--- a/lib/coverband/utils/results.rb
+++ b/lib/coverband/utils/results.rb
@@ -33,7 +33,7 @@ module Coverband
         return 100.0 unless eager_file
 
         runtime_relavant_lines = eager_file.relevant_lines - eager_file.covered_lines_count
-        runtime_file.runtime_relavant_calculations(runtime_relavant_lines, &:formatted_covered_percent)
+        runtime_file.runtime_relavant_calculations(runtime_relavant_lines) { |file| file.formatted_covered_percent }
       end
 
       ###
--- a/lib/coverband/utils/results.rb
+++ b/lib/coverband/utils/results.rb
@@ -33,7 +33,7 @@ module Coverband
         return 100.0 unless eager_file
 
         runtime_relavant_lines = eager_file.relevant_lines - eager_file.covered_lines_count
-        runtime_file.runtime_relavant_calculations(runtime_relavant_lines) { |file| file.formatted_covered_percent }
+        runtime_file.runtime_relavant_calculations(runtime_relavant_lines, &:formatted_covered_percent)
       end
 
       ###
--- a/lib/coverband/utils/results.rb
+++ b/lib/coverband/utils/results.rb
@@ -35,7 +35,7 @@ module Coverband
       end
 
       def runtime_relavent_lines(source_file)
-        return 0 unless eager_loading_coverage && runtime_coverage
+        return 0 unless runtime_coverage
 
         eager_file = get_eager_file(source_file)
         runtime_file = get_runtime_file(source_file)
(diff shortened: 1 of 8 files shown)


lib/coverband/utils/results.rb
```diff
@@ -20,6 +20,22 @@ def file_with_type(source_file, results_type)
...
+        eager_loading_coverage = get_results(Coverband::EAGER_TYPE)
+        runtime_coverage = get_results(Coverband::RUNTIME_TYPE)
+        return unless eager_loading_coverage && runtime_coverage
+
+        eager_file = eager_loading_coverage.source_files.find { |file| file.filename == source_file.filename }
+        runtime_file = runtime_coverage.source_files.find { |file| file.filename == source_file.filename }
+
+        return 0.0 unless runtime_file
+
+        return 100.0 unless eager_file
```
kbaum: Is this saying that whenever there is no eager coverage, return 100% runtime coverage? What if all the runtime lines are not hit? I am probably misunderstanding.

lib/coverband/utils/results.rb
```diff
@@ -20,6 +20,22 @@ def file_with_type(source_file, results_type)
...
+
+        runtime_relavant_lines = eager_file.relevant_lines - eager_file.covered_lines_count
```
kbaum: Very smart.

Seems like a small naming inconsistency in that `relevant_lines` is not `relevant_lines_count` like `covered_lines_count`?

lib/coverband/utils/results.rb
```diff
@@ -20,6 +20,22 @@ def file_with_type(source_file, results_type)
...
+        runtime_file.runtime_relavant_calculations(runtime_relavant_lines, &:formatted_covered_percent)
```
kbaum: I am having a hard time understanding this part. Could be because I am not as familiar with this area of the codebase. Seems like we are passing in a proc as the second argument but it's unclear what is passed into the proc unless we dive into the runtime_relavant_calculations implementation. Feels a little advanced to me but maybe it's needed?

lib/coverband/utils/source_file.rb
```diff
@@ -103,6 +104,13 @@ def initialize(filename, file_data)
         end
       end
 
+      def runtime_relavant_calculations(runtime_relavant_lines)
```
kbaum: Ok. Seems like the reason for the proc is that temporarily while this method is being called, relevant_lines behavior changes to return the instance variable? I am a little confused still because it does not seem to be passing in a proc over [here](https://github.com/danmayer/coverband/pull/270/files#diff-db957067981a31c0b4a7f4bcc0f8843aR42).

kbaum: Oh I see. The erb is invoking the `result#runtime_relevant_coverage` since it does not take a proc.

lib/coverband/utils/source_file.rb
```diff
@@ -171,7 +179,7 @@ def lines_strength
       end
 
       def relevant_lines
-        lines.size - never_lines.size - skipped_lines.size
+        @runtime_relavant_lines || (lines.size - never_lines.size - skipped_lines.size)
```
kbaum: I apologize as I am not as familiar with this area of the codebase.. would it be possible to make a `runtime_relevant_lines` method and instead invoke this method from the erb when needed?

danmayer: I don't think so because what we are actually calling is formatted coverage percentage, which in turn calls a number of other methods, this is the root of what changes the values of the calculations, but isn't the actual method... basically, the other approach requires having runtime_relavent_lines, runtime_relavent_coverage, and formatted_runtime_coverage and perhaps more

lib/coverband/utils/results.rb
danmayer: ha check git history I wrote this differently and rubocop yelled at me, so I actually think the original implementation is more readable if you agree I will go back to that and disable the robucop warning.

lib/coverband/utils/results.rb
danmayer: true this should just be return the normal runtime.percentage not calculating the runtime_relavant_lines

lib/coverband/utils/results.rb
kbaum: Agree that the original implementation was more readable.

lib/coverband/utils/source_file.rb
kbaum: Ok makes sense. Do you think that having all of those extra methods might be worth it at some point? I think it might make the code more straight forward.

danmayer: I think it would make things much more complicated because if you need to change any of the calculations or have a bug in one you now need to fix it in the original implementation and the duplicated methods. They would all be exactly the same except for using this base data.

I generally believe every line of code has a maintenance cost and opportunity to add bugs so minimizing lines of code by using slightly more complicated abstractions feels better to me.

lib/coverband/utils/results.rb
danmayer: ok cool reverted back to that implementation and updated rubocop to not complain about that.

lib/coverband/utils/source_file.rb
kbaum: Makes sense.

lib/coverband/utils/results.rb
danmayer: ok fixed this and added tests for all 3 cases

danmayer:
OK, I think this is ready for final review @kbaum

danmayer:
The file page shows the number of relevant lines total and runtime, but it isn't clear on a line by line basis what is what... which could be interesting, but I am not adding for now we can see if that feels valuable.

Screen Shot 2019-05-27 at 2 19 08 PM

lib/coverband/utils/results.rb
```diff
@@ -20,6 +20,31 @@ def file_with_type(source_file, results_type)
...
+        return 0.0 unless runtime_file
+
+        return runtime_file.formatted_covered_percent unless eager_file
+
+        runtime_relavant_lines = eager_file.relevant_lines - eager_file.covered_lines_count
+        runtime_file.runtime_relavant_calculations(runtime_relavant_lines) { |file| file.formatted_covered_percent }
+      end
+
+      def runtime_relavent_lines(source_file)
+        return 0 unless eager_loading_coverage && runtime_coverage
```
kbaum: Would it be possible that a file is only accessed at runtime? In this case would we just return runtime_coverage?

danmayer: yeah see line 43, this just expects that we will get eager and runtime coverage reports, which we should always expect so long as it has been initalized, but we check for the specific files down below.

https://github.com/danmayer/coverband/pull/270/files#diff-853e3c9a7d0f6b56a991f069a252d0cfR43

kbaum: I see. Although probably unlikely I feel like eager_loading_coverage could be nil if we are not in a rails project or using bundler. Assuming that is true, In this case, would we want to return `runtime_file.covered_lines_count`?

danmayer: OK, I thought how we save it would still have an empty report not nil, but I updated to use runtime regardless.

PR #274, opened 2019-05-28
List gems with no coverage

kbaum:
image

### DIFF
--- a/coverband.gemspec
+++ b/coverband.gemspec
@@ -35,6 +35,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'rake'
   spec.add_development_dependency 'resque'
   spec.add_development_dependency 'rubocop'
+  spec.add_development_dependency 'selenium-webdriver'
   # temporarily needed to run tests for classifier-reborn as part of benchmarks
   spec.add_development_dependency 'minitest-reporters'
 
--- a/coverband.gemspec
+++ b/coverband.gemspec
@@ -36,6 +36,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'resque'
   spec.add_development_dependency 'rubocop'
   spec.add_development_dependency 'selenium-webdriver'
+  spec.add_development_dependency 'webdrivers', '~> 3.0'
   # temporarily needed to run tests for classifier-reborn as part of benchmarks
   spec.add_development_dependency 'minitest-reporters'
 
--- a/lib/coverband/collectors/coverage.rb
+++ b/lib/coverband/collectors/coverage.rb
@@ -38,10 +38,10 @@ module Coverband
         @store.type = Coverband::EAGER_TYPE
       end
 
-      def eager_loading(&block)
+      def eager_loading
         old_coverage_type = @store.type
         eager_loading!
-        block.call
+        yield
       ensure
         report_coverage
         @store.type = old_coverage_type
--- a/lib/coverband/reporters/base.rb
+++ b/lib/coverband/reporters/base.rb
@@ -27,12 +27,13 @@ module Coverband
         ###
         def fix_reports(reports)
           # list all files, even if not tracked by Coverband (0% coverage)
-          tracked_glob = "#{Coverband.configuration.current_root}/{app,lib,config}/**/*.{rb}"
+          file_patterns = ["#{Coverband.configuration.current_root}/{app,lib,config}/**/*.{rb}"]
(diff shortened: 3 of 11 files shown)


coverband.gemspec
```diff
@@ -36,6 +36,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'resque'
   spec.add_development_dependency 'rubocop'
   spec.add_development_dependency 'selenium-webdriver'
+  spec.add_development_dependency 'webdrivers', '~> 3.0'
```
kbaum: @danmayer needed to get chrome headless working on travis.

danmayer:
on your screenshot I see the ungrouped tab, whenever that has happened to me it generally means there was a bug related to gems? Was that just while doing dev if you clear your Coverband Redis and go through the flow again does everything work?

lib/coverband/collectors/coverage.rb
```diff
@@ -38,10 +38,10 @@ def eager_loading!
         @store.type = Coverband::EAGER_TYPE
       end
 
-      def eager_loading(&block)
+      def eager_loading
         old_coverage_type = @store.type
         eager_loading!
-        block.call
+        yield
```
danmayer: ha I had done this on my branch because of rubocop ;)

lib/coverband/reporters/base.rb
```diff
@@ -27,12 +27,16 @@ def report(store, _options = {})
...
+          if Coverband.configuration.track_gems
+            file_patterns.concat(Gem.loaded_specs.values.map(&:full_require_paths)
+              .flatten.map { |path| "#{path}/**/*.{rb}" })
+          end
           filtered_report_files = {}
 
           reports.each_pair do |report_name, report_data|
             filtered_report_files[report_name] = {}
-            report_files = Coverband::Utils::Result.add_not_loaded_files(report_data, tracked_glob)
+            report_files = Coverband::Utils::Result.add_not_loaded_files(report_data, file_patterns)
```
danmayer: so this won't just list the gem with 0% but also all the files under the gem for gem details correct?

lib/coverband/utils/source_file.rb
```diff
@@ -242,7 +242,7 @@ def gem?
       end
 
       def gem_name
-        gem? ? short_name.split('/').first.gsub(%r{^\.}, '') : nil
+        gem? ? short_name.split('/').first.gsub(/^\./, '') : nil
```
danmayer: this is fine, but I did disable this rubocop warning as the %r is better in some cases when there are slashes etc.

test/benchmarks/benchmark.rake
```diff
@@ -130,7 +127,7 @@ namespace :benchmarks do
       x.config(time: 12, warmup: 5, suite: suite)
       x.report 'coverband' do
         work
-        Coverband.report_coverage(true)
+        Coverband.report_coverage
```
danmayer: ha fixed on my branch as well

danmayer:
OK take a look at that ungrouped thing after that looks good

kbaum:
Good catch. I definitely get an ungrouped tab:
image

danmayer:
Guessing it is adding files after the grouping step... Or they are added but for some reason not grouped...

danmayer:
Let me know if you want me to pull the branch and take a look at what is going on.

kbaum:
Sure! I might not have time to look at it until this weekend.

kbaum:
Noticed that the did_you_mean gem is being filtered out from gems due to the filter `"/Users/karlbaum/.gem/ruby/2.4.4"` in this method:
```ruby
      def gem_files(name, filter)
        grouped_gems = @files.select { |source_file| source_file.filename =~ /#{filter}/ }.group_by(&:gem_name)
        gem_lists = grouped_gems.values.map { |gem_files| Coverband::Utils::FileList.new(gem_files) }
        @grouped[name] = Coverband::Utils::GemList.new(gem_lists) if gem_lists.flatten.any?
        gem_lists
      end
```
The sourcefile path looks like `"/Users/karlbaum/.rubies/ruby-2.4.4/lib/ruby/gems/2.4.0/gems/did_you_mean-1.1.0/lib/did_you_mean.rb"`

Once it is filtered out in this method, it ends up in the other_files collection which puts it in "Ungrouped".

danmayer:
cool looks good if you get specs passing merge it in @kbaum

danmayer:
ahh it was just a rubocop warning... fixed the spacing waiting for tests to pass

Issue #272, opened 2019-05-27
clear runs reporting


danmayer:
I thought we fixed this but when I run `bundle exec rake coverband:clear` on the coverband demo app I get the below output.

```
bundle exec rake coverband:clear
D, [2019-05-27T15:10:53.311865 #24757] DEBUG -- : Coverband: Starting background reporting
D, [2019-05-27T15:10:53.375899 #24757] DEBUG -- : Coverband: Reported coverage via thread. Sleeping 30s
```

danmayer:
any ideas, since I think you noticed and resolved last time

kbaum:
Did some investigation. Coverband is being required when rails is required within the Rakefile. Coverband automatically invokes start and background thread is default true.

```ruby
  def self.start
    Coverband::Collectors::Coverage.instance
    # TODO: Railtie sets up at_exit after forks, via middleware, perhaps this hsould be
    # added if not rails or if rails but not rackserverrunning
    AtExit.register
    Background.start if configuration.background_reporting_enabled && !RackServerCheck.running?
  end
```
Kind of a bummer because I like the solution we tried here. Not sure of the best fix yet. 🤔

danmayer:
I have a path forward to at least resolve for the next release as we explore other options... will post PR shortly

danmayer:
resolved

PR #275, opened 2019-05-29
fix issues where coverband rake tasks record coverage issue #272

danmayer:
OK, so this prevents the rake tasks from reporting...

```
[master][~/projects/coverband_demo] bundle exec rake coverband:clear
[master][~/projects/coverband_demo]
```

I don't love the solution, but at least the test captures the flaw and that it is resolved. I am hoping we can still evolve this to something more intelligent.

### DIFF
--- a/lib/coverband.rb
+++ b/lib/coverband.rb
@@ -73,9 +73,8 @@ module Coverband
   end
 
   def self.tasks_to_ignore?
-    (defined?(Rake) &&
-    Rake.respond_to?(:application) &&
-    (Rake&.application&.top_level_tasks || []).any? { |task| Coverband::Configuration::IGNORE_TASKS.include?(task) })
+    defined?(Rake) &&
+      (Coverband::Configuration::IGNORE_TASKS & Array(Rake.application.top_level_tasks)).any?
   end
 
   def self.eager_loading_coverage!
--- a/lib/coverband.rb
+++ b/lib/coverband.rb
@@ -66,10 +66,16 @@ module Coverband
 
   def self.start
     Coverband::Collectors::Coverage.instance
-    # TODO: Railtie sets up at_exit after forks, via middleware, perhaps this hsould be
+    # TODO: Railtie sets up at_exit after forks, via middleware, perhaps this should be
     # added if not rails or if rails but not rackserverrunning
-    AtExit.register
-    Background.start if configuration.background_reporting_enabled && !RackServerCheck.running?
+    AtExit.register unless tasks_to_ignore?
+    Background.start if configuration.background_reporting_enabled && !RackServerCheck.running? && !tasks_to_ignore?
+  end
+
+  def self.tasks_to_ignore?
+    (defined?(Rake) &&
+    Rake.respond_to?(:application) &&
+    (Rake&.application&.top_level_tasks || []).any? { |task| Coverband::Configuration::IGNORE_TASKS.include?(task) })
   end
 
   def self.eager_loading_coverage!
--- a/lib/coverband.rb
+++ b/lib/coverband.rb
@@ -73,8 +73,9 @@ module Coverband
   end
 
   def self.tasks_to_ignore?
-    defined?(Rake) &&
-      (Coverband::Configuration::IGNORE_TASKS & Array(Rake.application.top_level_tasks)).any?
+    (defined?(Rake) &&
+    Rake.respond_to?(:application) &&
+    (Rake&.application&.top_level_tasks || []).any? { |task| Coverband::Configuration::IGNORE_TASKS.include?(task) })
   end
 
   def self.eager_loading_coverage!
--- a/lib/coverband/configuration.rb
+++ b/lib/coverband/configuration.rb
@@ -19,9 +19,7 @@ module Coverband
     # * should we skip /bin/rails webpacker:compile ?
     # * Perhaps detect heroku deployment ENV var opposed to tasks?
     #####
-    IGNORE_TASKS = ['assets:clean',
-                    'assets:precompile',
-                    'coverband:clear',
+    IGNORE_TASKS = ['coverband:clear',
                     'coverband:coverage',
                     'coverband:coverage_server',
                     'coverband:migrate']
--- a/lib/coverband/configuration.rb
+++ b/lib/coverband/configuration.rb
@@ -12,6 +12,20 @@ module Coverband
     attr_writer :logger, :s3_region, :s3_bucket, :s3_access_key_id, :s3_secret_access_key
     attr_reader :track_gems, :ignore, :use_oneshot_lines_coverage
 
+    #####
+    # TODO: This is is brittle and not a great solution to avoid deploy time
+    # actions polluting the 'runtime' metrics
+    #
+    # * should we skip /bin/rails webpacker:compile ?
+    # * Perhaps detect heroku deployment ENV var opposed to tasks?
+    #####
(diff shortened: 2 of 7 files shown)


lib/coverband.rb
```diff
@@ -66,10 +66,16 @@ def self.configuration
...
     # added if not rails or if rails but not rackserverrunning
-    AtExit.register
-    Background.start if configuration.background_reporting_enabled && !RackServerCheck.running?
+    AtExit.register unless tasks_to_ignore?
+    Background.start if configuration.background_reporting_enabled && !RackServerCheck.running? && !tasks_to_ignore?
+  end
+
+  def self.tasks_to_ignore?
+    (defined?(Rake) &&
+    Rake.respond_to?(:application) &&
```
kbaum: Don't think we need this check. Seems like it's always had an application. This commit moves the method to a different file in 2009 but it was already there at that point:
https://github.com/ruby/rake/commit/2de58bc9e1b1ab89564f160c58f768cefa736e70

lib/coverband.rb
```diff
@@ -66,10 +66,16 @@ def self.configuration
...
+    (Rake&.application&.top_level_tasks || []).any? { |task| Coverband::Configuration::IGNORE_TASKS.include?(task) })
```
kbaum: We could also assume that top_level_tasks is always present since before 2009: https://github.com/ruby/rake/commit/2de58bc9e1b1ab89564f160c58f768cefa736e70#diff-26729347a9c3b31c5c9f4305f965d8e3

That could take this down to?
```ruby
defined?(Rake) &&
  (Coverband::Configuration::IGNORE_TASKS & Array(Rake.application.top_level_tasks)).any?
```

lib/coverband/configuration.rb
```diff
@@ -12,6 +12,20 @@ class Configuration
...
     attr_reader :track_gems, :ignore, :use_oneshot_lines_coverage
 
+    #####
+    # TODO: This is is brittle and not a great solution to avoid deploy time
+    # actions polluting the 'runtime' metrics
+    #
+    # * should we skip /bin/rails webpacker:compile ?
+    # * Perhaps detect heroku deployment ENV var opposed to tasks?
+    #####
+    IGNORE_TASKS = ['assets:clean',
```
kbaum: I don't think we need `assets:clean` or `assets:precompile` anymore. Ideally we are only ignoring only `coverband:` namespaced tasks.

test/forked/rails_rake_full_stack_test.rb
```diff
@@ -17,4 +17,16 @@ class RailsRakeFullStackTest < Minitest::Test
     pundit_coverage = store.coverage[pundit_file]
     assert_nil pundit_coverage
   end
+
+  test "ignored rake tasks don't add coverage" do
+    store.instance_variable_set(:@redis_namespace, 'coverband_test')
```
kbaum: Can this be moved to test_helper possibly? Also, does this work since we are shelling out on line 24?

danmayer: it does, it is weird... basically, we shell out and the shelled out code used the dummy rails `config/coverband.rb` which sets the namespace in the shell... this line sets the namespace for the code not shelled out where we call `store` directly to check the data... If we don't set it here we are checking against a different redis namespace.

lib/coverband/configuration.rb
danmayer: I will test that on coverband demo and see if we can remove

lib/coverband.rb
danmayer: Wow, nice digging in Rake source... This is way simpler, updating.

lib/coverband/configuration.rb
kbaum: Ok sounds good. If we do need it, it would be a good opportunity to add a test to expose that need if we don't have one already.

test/forked/rails_rake_full_stack_test.rb
kbaum: Ok. Should we just globally set our namespace to coverband_test in test_helper?

lib/coverband.rb
danmayer: hmm when trying to test removing `assets:clean` on heroku this blew up...

I think when you run rake these are all there but when you run other Ruby processes there is stuff that tries to patch Rake or it is not fully loaded. Adding the checks back.

```
 ! Unable to load application: NoMethodError: undefined method `application' for Rake:Module
2019-05-29T14:20:39.755959+00:00 app[web.1]: bundler: failed to load command: puma (/app/vendor/bundle/ruby/2.6.0/bin/puma)
2019-05-29T14:20:39.756036+00:00 app[web.1]: NoMethodError: undefined method `application' for Rake:Module
2019-05-29T14:20:39.756039+00:00 app[web.1]: /app/vendor/bundle/ruby/2.6.0/bundler/gems/coverband-5c6f3d14b3b6/lib/coverband.rb:77:in `tasks_to_ignore?'
2019-05-29T14:20:39.756040+00:00 app[web.1]: /app/vendor/bundle/ruby/2.6.0/bundler/gems/coverband-5c6f3d14b3b6/lib/coverband.rb:71:in `start'
2019-05-29T14:20:39.756042+00:00 app[web.1]: /app/vendor/bundle/ruby/2.6.0/bundler/gems/coverband-5c6f3d14b3b6/lib/coverband.rb:99:in `<module:Coverband>'
2019-05-29T14:20:39.756043+00:00 app[web.1]: /app/vendor/bundle/ruby/2.6.0/bundler/gems/coverband-5c6f3d14b3b6/lib/coverband.rb:32:in `<main>'
2019-05-29T14:20:39.756045+00:00 app[web.1]: /app/vendor/bundle/ruby/2.6.0/gems/bootsnap-1.4.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:21:in `require'
2019-05-29T14:20:39.756047+00:00 app[web.1]: /app/vendor/bundle/ruby/2.6.0/gems/bootsnap-1.4.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:21:in `block in require_with_bootsnap_lfi'
2019-05-29T14:20:39.756048+00:00 app[web.1]: /app/vendor/bundle/ruby/2.6.0/gems/bootsnap-1.4.1/lib/bootsnap/load_path_cache/loaded_features_index.rb:83:in `register'
2019-05-29T14:20:39.756050+00:00 app[web.1]: /app/vendor/bundle/ruby/2.6.0/gems/bootsnap-1.4.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:20:in `require_with_bootsnap_lfi'
2019-05-29T14:20:39.756052+00:00 app[web.1]: /app/vendor/bundle/ruby/2.6.0/gems/bootsnap-1.4.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:29:in `require'
```

lib/coverband/configuration.rb
danmayer: yes no longer needed removing.

danmayer:
let me know if there is anything else @kbaum

test/forked/rails_rake_full_stack_test.rb
danmayer: I had tried that, not exactly sure why it wasn't always getting picked up... See the settings in the helper...

https://github.com/danmayer/coverband/blob/master/test/test_helper.rb#L32

Any ideas why this would be in a state where the store doesn't have the correct namespace @kbaum ?

PR #278, opened 2019-05-30
Revert "List gems with no coverage"

kbaum:
Reverts danmayer/coverband#274


danmayer:
makes sense for now, thanks merging. I will get a RC 3 out later today and get it testing on some apps

PR #279, opened 2019-06-07
Remove unneeded dup

kbaum:
Looks like ::Coverage.peek_result makes a new hash each call.

```
irb(main):002:0> Coverage.peek_result.object_id == Coverage.peek_result.object_id
=> false
```

image


danmayer:
good call thanks

PR #282, opened 2019-06-08
Remove unnecessary get_report abstraction

kbaum:


danmayer:
good call feels like that had some purpose in the past but definitely just extra work for no reason now. Thanks

PR #283, opened 2019-06-11
Remove repeated integrations/background require from coverband.rb

yuriyalekseyev:
Hi,
Just noticed that `require 'coverband/integrations/background'` was called two times.


kbaum:
Thanks!

danmayer:
thanks @yuriyalekseyev

PR #286, opened 2019-06-13
Remove unused link from table of contents in README

yuriyalekseyev:
Another micro-fix from me:
"Verify correct installation" section was removed in c26f379, but a link in table-of-contents stayed since then.


danmayer:
Thanks always good to keep the documentation up to date. Appreciate it @yuriyalekseyev

PR #287, opened 2019-06-13
Don't clobber environment method on main

shioyama:
The `environment` method in the Coverband rake task is currently defined on the global namespace, see:

https://kevinjalbert.com/defined_methods-in-rake-tasks-you-re-gonna-have-a-bad-time/

We had an issue because a factory was defining an attribute named `environment` and hitting a conflict with this globally-defined method.

Instead of defining a method here, how about a proc in a closure? This keeps the code encapsulated and will avoid conflicts like this.

### DIFF
--- a/lib/coverband/utils/tasks.rb
+++ b/lib/coverband/utils/tasks.rb
@@ -4,7 +4,7 @@ namespace :coverband do
   # handles configuring in require => false and COVERBAND_DISABLE_AUTO_START cases
   Coverband.configure unless Coverband.configured?
 
-  environment = -> {
+  environment = lambda {
     Coverband.configuration.report_on_exit = false
     Coverband.configuration.background_reporting_enabled = false
     Rake.application['environment'].invoke if Rake::Task.task_defined?('environment')
--- a/lib/coverband/utils/tasks.rb
+++ b/lib/coverband/utils/tasks.rb
@@ -4,15 +4,15 @@ namespace :coverband do
   # handles configuring in require => false and COVERBAND_DISABLE_AUTO_START cases
   Coverband.configure unless Coverband.configured?
 
-  def environment
+  environment = -> {
     Coverband.configuration.report_on_exit = false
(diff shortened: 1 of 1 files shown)


shioyama:
p.s. there doesn't seem to be any test for the rake task so I didn't test that this works beyond a very simple check.

shioyama:
To see the issue with the current code, create a vanilla Rails app, run a console and first confirm that no `environment` method is defined on `main` after loading tasks:

```ruby
Rails.application.load_tasks
method(:environment)
#=> NameError (undefined method `environment'  for class ...
```

Now add coverband to the Gemfile and try again:

```ruby
Rails.application.load_tasks
method(:environment)
=> #<Method: main.environment>
```

You can see it's the one from the rake task:

```ruby
method(:environment).source_location
=> ["/home/chris/.rbenv/versions/2.6.1/lib/ruby/gems/2.6.0/gems/coverband-4.2.0/lib/coverband/utils/tasks.rb", 6]
```

kbaum:
This is a great catch and interesting solution that I wouldn't have thought of. I actually think that we can just remove this method as what it was trying to do is handled here by calling `tasks_to_ignore?`.

https://github.com/danmayer/coverband/blob/master/lib/coverband.rb#L71

danmayer:
Yeah, interesting and thanks for the fix @shioyama I believe as @kbaum says we can remove this... The only line I wasn't sure about was...

`Rake.application['environment'].invoke if Rake::Task.task_defined?('environment')` but I remember this is a legacy from when we took `coverband:baseline` it is no longer really needed to my knowledge.

I will confirm that and remove it.

danmayer:
Thanks again @shioyama I fixed this on the master branch here: https://github.com/danmayer/coverband/commit/a58c3bcab328ff6e4e93122e7c901f71b7619a65

It will go out in the next release.

shioyama:
Thanks for the quick response + fix! Glad I could help.

PR #290, opened 2019-06-19
Memory improvements

danmayer:
I believe all of these improvements should help resolve the memory issues seen in issue #277 

This fixes issues around adding oneshot
   * it had a large memory hit due to `::Coverage.line_stub(file)`
   * bugs related to previous coverage
   * gets to 0 retained memory between requests
   * seems to have NO memory growth flatlining at basically the same as non-coverband
   * has slightly increased up-front allocations to avoid later allocations (stubs cache, etc)
   * this is recommended for anyone running 2.6, and likely will become the default after being optional in this release

Adds support for simulated_oneshot for older Ruby versions that don't support it
   * gets to 0 retained memory between requests
   * seems to significantly reduce Heroku memory growth beyond the fixes of our current implementation
   * optional configurable opt-in: `config.simulate_oneshot_lines_coverage = true`

Beyond that, it includes some other fixes that improve the memory usage on the current default as well... Which when running on Heroku shows significantly reduced memory growth. Such as the `@@path_cache = {}` which avoids repeatedly making shortened path strings.

benchmark now supports running in 3 modes... Working on the same for the forked tests

* default: `bundle exec rake benchmarks:run_big`
* oneshot:  `ONESHOT=true bundle exec rake benchmarks:run_big`
* `SIMULATE_ONESHOT=true bundle exec rake benchmarks:run_big`

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -29,7 +29,8 @@ end
 
 desc 'load irb with this gem'
 task :console do
-  exec 'irb -I lib -r coverband'
+  puts 'running console'
+  exec 'bundle exec irb -I lib -r coverband'
 end
 
 # This is really just for testing and development because without configuration
--- a/Rakefile
+++ b/Rakefile
@@ -30,7 +30,7 @@ end
 desc 'load irb with this gem'
 task :console do
   puts 'running console'
-  exec 'bundle exec irb -I lib -r coverband'
+  exec 'bundle exec console'
 end
 
 # This is really just for testing and development because without configuration
--- a/lib/coverband/adapters/base.rb
+++ b/lib/coverband/adapters/base.rb
@@ -22,7 +22,7 @@ module Coverband
         raise ABSTRACT_KEY
       end
 
-      def clear_file!(_file)
+      def clear_file!(_file = nil)
         raise ABSTRACT_KEY
       end
 
--- a/lib/coverband/adapters/base.rb
+++ b/lib/coverband/adapters/base.rb
@@ -131,7 +131,11 @@ module Coverband
       end
 
       def array_add(latest, original)
-        latest.map.with_index { |v, i| (v && original[i]) ? v + original[i] : nil }
+        if Coverband.configuration.use_oneshot_lines_coverage
+          latest.map.with_index { |v, i| (v + original[i] >= 1 ? 1 : 0) if v && original[i] }
+        else
+          latest.map.with_index { |v, i| (v && original[i]) ? v + original[i] : nil }
+        end
       end
     end
   end
--- a/lib/coverband/adapters/base.rb
+++ b/lib/coverband/adapters/base.rb
@@ -15,6 +15,7 @@ module Coverband
 
       def initialize
         @file_hash_cache = {}
+        @type = Coverband::RUNTIME_TYPE
       end
 
       def clear!
@@ -100,6 +101,10 @@ module Coverband
       end
 
       def merge_reports(new_report, old_report, options = {})
+        # transparently update from RUNTIME_TYPE = nil to RUNTIME_TYPE = :runtime
+        # transparent update for format coveband_3_2
+        old_report = coverage(nil) if old_report.nil? && type == Coverband::RUNTIME_TYPE
+
         new_report = expand_report(new_report) unless options[:skip_expansion]
         keys = (new_report.keys + old_report.keys).uniq
         keys.each do |file|
--- a/lib/coverband/adapters/base.rb
+++ b/lib/coverband/adapters/base.rb
@@ -103,7 +103,7 @@ module Coverband
       def merge_reports(new_report, old_report, options = {})
         # transparently update from RUNTIME_TYPE = nil to RUNTIME_TYPE = :runtime
         # transparent update for format coveband_3_2
-        old_report = coverage(nil) if old_report.nil? && type == Coverband::RUNTIME_TYPE
+        old_report = coverage(nil, override_type: nil) if old_report.nil? && type == Coverband::RUNTIME_TYPE
 
         new_report = expand_report(new_report) unless options[:skip_expansion]
         keys = (new_report.keys + old_report.keys).uniq
--- a/lib/coverband/adapters/base.rb
+++ b/lib/coverband/adapters/base.rb
@@ -69,8 +69,21 @@ module Coverband
 
       def split_coverage(types)
         types.reduce({}) do |data, type|
-          data.update(type => coverage(type))
+          if type == Coverband::RUNTIME_TYPE && Coverband.configuration.simulate_oneshot_lines_coverage
+            data.update(type => simulated_runtime_coverage)
+          else
+            data.update(type => coverage(type))
+          end
         end
+      end
+
+      def simulated_runtime_coverage
+        runtime_data = coverage(Coverband::RUNTIME_TYPE)
+        eager_data = coverage(Coverband::EAGER_TYPE)
+        eager_data.each_pair do |_key, vals|
+          vals['data'].map! { |el| el ? (0 - el) : el }
+        end
+        merge_reports(runtime_data, eager_data, skip_expansion: true)
       end
 
       def merged_coverage(types)
--- a/lib/coverband/adapters/base.rb
+++ b/lib/coverband/adapters/base.rb
@@ -80,8 +80,8 @@ module Coverband
       def simulated_runtime_coverage
         runtime_data = coverage(Coverband::RUNTIME_TYPE)
         eager_data = coverage(Coverband::EAGER_TYPE)
-        eager_data.each_pair do |_key, vals|
-          vals['data'].map! { |el| el ? (0 - el) : el }
+        eager_data.values do |vals|
+          vals['data'].map! { |line_coverage| line_coverage ? (0 - line_coverage) : line_coverage }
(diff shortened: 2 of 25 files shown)


danmayer:
I am going to leave oneshot running for a while and try to improve the test coverage around the various modes. All 3 modes now perform much better than the currently live release.

Screen Shot 2019-06-19 at 12 33 49 PM

lib/coverband/utils/file_path_helper.rb
```diff
@@ -8,16 +8,23 @@ module Utils
...
       # Takes a full path and converts to a relative path
       ###
       def full_path_to_relative(full_path)
+        return @@path_cache[full_path] if @@path_cache.key?(full_path)
+
         relative_filename = full_path
-        Coverband.configuration.all_root_paths.each do |root|
-          relative_filename = relative_filename.gsub(/^#{root}/, './')
+        Coverband.configuration.all_root_patterns.each do |root|
+          relative_filename = relative_filename.sub(root, './')
```
kbaum: I wonder if it would help memory to use `sub!`

lib/coverband/configuration.rb
```diff
@@ -7,7 +7,8 @@ class Configuration
                   :reporter, :redis_namespace, :redis_ttl,
                   :background_reporting_enabled,
                   :background_reporting_sleep_seconds, :test_env,
-                  :web_enable_clear, :gem_details, :web_debug, :report_on_exit
+                  :web_enable_clear, :gem_details, :web_debug, :report_on_exit,
+                  :simulate_oneshot_lines_coverage
```
kbaum: Could we instead just use the configuration `use_oneshot_lines_coverage` and then only simulate if ruby version does not support one shot line coverage?

lib/coverband/adapters/base.rb
```diff
@@ -124,8 +130,15 @@ def merge_expanded_data(new_expanded, old_expanded)
         }
       end
 
+      # TODO: This should only be 2 cases get our dup / not dups aligned
       def array_add(latest, original)
```
danmayer: throughout our code we now have code paths that either have a froze or not frozen coverage... this creates problems... in the end we probably want to pick ONE way of going starting with a full set and using in memory `!` or starting with frozen and then always reducing to new....

lib/coverband/utils/file_path_helper.rb
danmayer: yeah probably would... see this line https://github.com/danmayer/coverband/pull/290/files#diff-2d63c89a82b59b5370220c4697cdbee2R133

basically, I think we have a larger cleanup around all in memory edits or all reduction edits...

lib/coverband/configuration.rb
danmayer: yeah I did consider that, but there are some tradeoffs, so I wanted to make it a bit more explicit... basically while it doesn't store anything in memory between requests it does have a bit higher allocations.

danmayer:
these now  pass

`ONESHOT=true bundle exec rake forked_tests`
`SIMULATE_ONESHOT=true bundle exec rake forked_tests`

lib/coverband/configuration.rb
kbaum: One thought. Was under the impression that peek_results is what causes the memory bloat. If that is the case, why would simulate_oneshot_coverage help since we are still calling peek_results?

lib/coverband/adapters/base.rb
```diff
@@ -124,8 +130,15 @@ def merge_expanded_data(new_expanded, old_expanded)
...
-        latest.map.with_index { |v, i| (v && original[i]) ? v + original[i] : nil }
+        if Coverband.configuration.use_oneshot_lines_coverage
+          latest.map!.with_index { |v, i| (v + original[i] >= 1 ? 1 : 0) if v && original[i] }
```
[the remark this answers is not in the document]
danmayer: yeah I thought of this, but I think it ends up a bit more complex...

```
if Coverband.configuration.use_oneshot_lines_coverage
          latest.map!.with_index { |v, i| (v + original[i] >= 1 ? 1 : 0) if v && original[i] }
        elsif Coverband.configuration.simulate_oneshot_lines_coverage
          latest.map.with_index { |v, i| (v + original[i] >= 1 ? 1 : 0) if v && original[i] }
```

becomes

```
if (Coverband.configuration.use_oneshot_lines_coverage || Coverband.configuration.simulate_oneshot_lines_coverage)
  if latest.frozen?
       latest.map.with_index { |v, i| (v + original[i] >= 1 ? 1 : 0) if v && original[i] }
  else
       latest.map!.with_index { |v, i| (v + original[i] >= 1 ? 1 : 0) if v && original[i] }
```

I would like to update the code so we have a single strategy all the way through, but it will be a bit of a bigger effort and I am hoping to combine that with dropping support for older Ruby etc and do a full cleanup pass.

Thinking we always get the hash from `Coverage` and try to filter it early to a isolated object, then use all in memory reductions beyond the delta class.

lib/coverband/configuration.rb
danmayer: It removes storing any memory between requests... so there is no `@@previous_coverage` holding a large JSON structure... Basically, those were the two biggest issues

*  `peek_results` itself can cause memory bloat if you end up in a base situation when we have to malloc and GC memory all the time... Can't fix that until `2.6.0`
* We significantly increase memory all the time if we have to keep `@@previous_coverage` in memory between requests because then we can't even GC it. (simulated oneshot or real oneshot, fixes this issue)

I guess the 3rd issue is what you have been trying to fix, which is that our serialization just has far to large of memory cost.

Make sense @kbaum ?

danmayer:
I think this is ready to merge any final thoughts @kbaum

lib/coverband/configuration.rb
kbaum: Makes sense. Thanks for explanation!

Rakefile
```diff
@@ -29,7 +29,8 @@ end
 
 desc 'load irb with this gem'
 task :console do
-  exec 'irb -I lib -r coverband'
+  puts 'running console'
+  exec 'bundle exec irb -I lib -r coverband'
```
kbaum: Does this do something different then `bundle console`?

lib/coverband/adapters/base.rb
```diff
@@ -64,24 +69,34 @@ def covered_files
...
+          else
+            data.update(type => coverage(type))
+          end
         end
       end
 
+      def simulated_runtime_coverage
+        runtime_data = coverage(Coverband::RUNTIME_TYPE)
+        eager_data = coverage(Coverband::EAGER_TYPE)
+        eager_data.each_pair do |_key, vals|
```
kbaum: Could just call `eager_data.values` ?

lib/coverband/adapters/base.rb
```diff
@@ -64,24 +69,34 @@ def covered_files
...
+          vals['data'].map! { |el| el ? (0 - el) : el }
```
kbaum: Looks like el is the coverage for a line in a file. I think a more descriptive variable might be helpful when reading this code. Something like this?

```ruby
vals['data'].map! { |line_coverage| line_coverage ? (0 - line_coverage) : line_coverage }
```

lib/coverband/adapters/base.rb
```diff
@@ -64,24 +69,34 @@ def covered_files
...
-          data.update(type => coverage(type))
+          if type == Coverband::RUNTIME_TYPE && Coverband.configuration.simulate_oneshot_lines_coverage
+            data.update(type => simulated_runtime_coverage)
+          else
+            data.update(type => coverage(type))
+          end
         end
       end
 
+      def simulated_runtime_coverage
```
kbaum: This is clever. I think one limitation is for code that's hit during eager_loading and runtime. Something we might want to point out to users.

lib/coverband/adapters/base.rb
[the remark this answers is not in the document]
kbaum: Gotcha. By one strategy, do you mean always use oneshot?

Rakefile
danmayer: nope I hadn't used bundle console... I guess I could just delete this task ;)

lib/coverband/adapters/base.rb
danmayer: good call updated

Rakefile
danmayer: opposed to removing I updated to bundle console in the rake task cause I would forget again ;)

lib/coverband/adapters/base.rb
danmayer: done

lib/coverband/adapters/base.rb
danmayer: yeah, perhaps we should document that `simulated_runtime_coverage` has a nice advantage of overall reduced runtime memory usage, but it can't fully track eager vs runtime coverage and isn't as good as real oneshot for memory usage.

lib/coverband/adapters/base.rb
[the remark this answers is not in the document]
danmayer: In this case, I am more thinking of a strategy for how we process Coverage data... like having rules where we try to never dup and always filter in-memory vs into copies...

PR #292, opened 2019-06-21
Basic Auth

jjb:
Support for basic authentication.

If you like the idea, I can add tests and docs. Let me know what you think!

Thanks for a great project!

(based on 4.2.0)

### DIFF
--- a/lib/coverband/configuration.rb
+++ b/lib/coverband/configuration.rb
@@ -10,7 +10,7 @@ module Coverband
                   :background_reporting_sleep_seconds, :test_env,
                   :web_enable_clear, :gem_details, :web_debug
 
-    attr_writer :logger, :s3_region, :s3_bucket, :s3_access_key_id, :s3_secret_access_key
+    attr_writer :logger, :s3_region, :s3_bucket, :s3_access_key_id, :s3_secret_access_key, :password
     attr_reader :track_gems
 
     def initialize
@@ -38,6 +38,7 @@ module Coverband
       @gem_details = false
       @groups = {}
       @web_debug = false
+      @password = nil
 
       # TODO: should we push these to adapter configs
       @s3_region = nil
@@ -54,6 +55,10 @@ module Coverband
                   else
                     Logger.new(STDOUT)
                   end
+    end
+
+    def password
+      @password || ENV['COVERBAND_PASSWORD']
     end
 
     def s3_bucket
--- a/lib/coverband/reporters/web.rb
+++ b/lib/coverband/reporters/web.rb
@@ -17,8 +17,19 @@ module Coverband
                                    urls: [/.*\.css/, /.*\.js/, /.*\.gif/, /.*\.png/])
       end
 
+      def check_auth
+        return true unless Coverband.configuration.password
+        auth_header = request.get_header("HTTP_AUTHORIZATION")
+        return unless auth_header
+        Coverband.configuration.password == Base64.decode64((auth_header).split[1]).split(":")[1]
+      end
+
       def call(env)
         @request = Rack::Request.new(env)
+
+        unless check_auth
+          return [401, {"www-authenticate" => 'Basic realm=""'}, ['']]
+        end
 
         if request.post?
           case request.path_info


danmayer:
yeah we have mostly just recommended how to do this via rails route rules in the readme or in the demo example

https://github.com/danmayer/coverband_demo/blob/master/config/routes.rb#L10

I guess it would be fine to just implement it and include it as an option. 

Thanks I will resolve conflicts and get this merged.

jjb:
Cool!

I think the problem with the method you linked to is that if there is no user/pass header, there's no way to send back a 401 to ask the user for the user/pass (that I know of). So a full user/pass UX can't be implemented just with a route constraint.

danmayer:
good point, I will get this merged just working on another branch at the moment.

danmayer:
merged thanks

kopylovvlad:
Thank you for the work but there are problems.

The first one is `request.get_header` exist only in `rack` >= 2.0.0. If I have an application with rails 4.2 it doesn't work. It raises an exception "undefined method `get_header' for Rack::Request".
It seems to me that better way will be use `request.env['HTTP_AUTHORIZATION']` instead `request.get_header('HTTP_AUTHORIZATION')`.

The second one is why did you decided to use and check only passsword? It is a potential vulnerability. In my eyes, basic http auth should check username and password, and config should be something like that `config.auth = 'username:password'`. 

What do you think about it?

kbaum:
I'm not sure this kind of feature should be included in coverband. Feels like the library is getting outside of its core competency. Looking at a widely used gem like sidekiq and it looks like they don't include authentication but instead provide examples of how to do it:

https://github.com/mperham/sidekiq/wiki/Monitoring#authentication

kopylovvlad:
I would like to use `Rails HTTP Basic Auth from Routes` in my application. But I don't clearly understand how to implement it.
It will be good to write something like `Coverband::Reporters::Web.use Rack::Auth::Basic do |username, password|` but it doesn't have `.use` method

kbaum:
@kopylovvlad - i like the idea of allowing users to implement basic auth themselves as you suggested. Not sure why `Coverband::Reporters::Web.use` doesn't work but we should look into it and fix and then add an example to the wiki.

kopylovvlad:
It doesn't work because `Coverband::Reporters::Web` hasn't middleware support therefore it hasn't the method `use` 😢

jjb:
@kopylovvlad 

whoops, sorry about the <2 incompatibility 

> why did you decided to use and check only passsword?

because there's only one "account"/"user" so there's no reason to have a username

> It is a potential vulnerability

what's the vulnerability?


@kbaum 

> I'm not sure this kind of feature should be included in coverband. Feels like the library is getting outside of its core competency

I agree that if it were possible it would be better to do it at some other layer, but I think it's not possible. see the discussion before this was merged. "I think the problem with the method…" explains why the user can't implement it themselves. Let me know if you have ideas!

danmayer:
@kopylovvlad  before this PR was implemented we just recommended folks do it themselves there are dozens of ways to add auth around the middleware. The reporter is definitely middleware, so not sure what you mean that it isn't rack middleware, but it might not support all usages. If you have an example of something with `use`, we could look at supporting it.

This implementation does nothing to stop folks from implementing it themselves it is just a quick way for most folks to do it on a modern setup. That being said if we can test that `request.env['HTTP_AUTHORIZATION']` works in both old and new Rack, happy to take a PR switching. 

I am looking to drop support for Rails 4, I asked about this a bit ago in the issues and haven't heard much from folks still on Rails 4. So I am not sure I want to change things too much to support a layer we are deprecating.  In most real apps, I have this protected by devise router helpers...

You can still see at least one alternative example in the readme:

https://github.com/danmayer/coverband#coverage-report

If you look at the readme history you will see an evolution of 3 or 4 different options for authentication. Folks kept sending PRs with there latest / greatest version. I guess we could have a wiki page with a big list of alternative options. 

Another example that you could do is in the coverband_demo routes

# defined
https://github.com/danmayer/coverband_demo/blob/master/config/routes.rb#L8
# usage
https://github.com/danmayer/coverband_demo/blob/master/config/routes.rb#L31

yeah @kbaum I can't decide, I like it is super easy to setup and want to avoid folks accidentally exposing sensitive source code...  Sidekiq doesn't normally expose such sensitive data... That being said it really isn't core to what we are trying to provide...

@kopylovvlad let me know if these alternatives work for you or you have a bit more info about what we would need to do to support `Rack.use`

kbaum:
> Sidekiq doesn't normally expose such sensitive data

That's a good point and I didn't think of that although even though it's not as sensitive, I cannot see a use case for exposing sidekiq admin without auth. My thought is that this change is not core to coverband and it does not seem to make basic auth that much easier. 

@jjb - would you be ok with using basic auth as defined in coverband demo [here](https://github.com/danmayer/coverband_demo/blob/master/config/routes.rb#L8)? Seems relatively straight forward to me. Let us know what you think.

jjb:
That method does not return 401 on failure. the user will not be prompted for a user/pass in the browser. That’s why I made this PR :)

PR #296, opened 2019-06-24
compress data in redis

jjb:
et me know what you think! i didn't try it yet. if you like the idea i'll try it and i'll squish the commits.

### DIFF
--- a/lib/coverband/adapters/redis_store.rb
+++ b/lib/coverband/adapters/redis_store.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'zlib'
+
 module Coverband
   module Adapters
     ###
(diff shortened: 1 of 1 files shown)


jjb:
for size do we want the size in the db or the size after uncompressed?

```ruby
      def size
        @redis.get(base_key).bytesize
      end
```

danmayer:
I guess since this doesn't effect runtime memory and adds a bit of CPU to runtime, not sure if it is something most people need... I have pretty large systems running and haven't seen Redis memory usage over 1.5mb

Was this driven by running into large usage of Redis space or another concern? Do you have a project that is having an issue related to this @jjb ?

It is a good idea, but I am trying to avoid adding things without strong support for the need.

jjb:
gotcha

for some reason i thought the data was bigger

i agree, no need to increase complexity

PR #298, opened 2019-06-26
Run big but generate coverage data

kbaum:
Generate coverage data for a more realistic benchmark.


danmayer:
awesome thanks! Merge when ready @kbaum

PR #304, opened 2019-07-26
Hash redis store using lua for writes

kbaum:


danmayer:
I will look more closely at this PR over the weekend

lib/coverband.rb
```diff
@@ -28,6 +29,8 @@
 require 'coverband/integrations/background_middleware'
 require 'coverband/integrations/rack_server_check'
 
+Coverband::Adapters::RedisStore = Coverband::Adapters::HashRedisStore
```
kbaum: @danmayer Going to revert this. I just did this so i can test the HashRedisStore with the entire test suite. Thinking we want to keep both the RedisStore and the HashRedisStore around a while?

kbaum:
Cleaned things up so this should be in a more mergeable state.

lib/coverband.rb
```diff
@@ -28,6 +29,8 @@
 require 'coverband/integrations/background_middleware'
 require 'coverband/integrations/rack_server_check'
 
+Coverband::Adapters::RedisStore = Coverband::Adapters::HashRedisStore if ENV['COVERBAND_HASH_REDIS_STORE']
```
kbaum: @danmayer - so we can run

```
COVERBAND_HASH_REDIS_STORE=t bundle exec rake test
```

And see how the HashRedisStore works with our test suite.

lib/coverband/adapters/hash_redis_store.rb
```diff
@@ -0,0 +1,164 @@
...
+
+      def save_report(report)
+        roots = Coverband.configuration.all_root_paths
+        report_time = Time.now.to_i
+        updated_time = type == Coverband::EAGER_TYPE ? nil : report_time
+        script_id = hash_incr_script
+        @redis.pipelined do
+          keys = report.map do |file, data|
+            relative_file = full_path_to_relative(file)
+            absolute_path = relative_path_to_full(file, roots)
```
kbaum: @danmayer - There's some refactoring i need to do with calculating file_hash. Feel like this conversion back and forth between relative and absolute files over and over is a little messy and probably not great for performance. Thinking about pushing this down to the file_hash method and caching the hash by relative file to avoid converting from relative to absolute file each time. Some work still to do there.

danmayer: yeah, good call I think if we worked on our reporter we could try to have everything work as a relative path and avoid all the back and forth. We still have to always convert the Coverage stdlib to relative path though.

danmayer:
Yeah, I think for at least a version or two we should offer both stores... I can run them on a few systems and try to get more real-world data on the tradeoffs / clear winner. Love having it as an optional first version.

lib/coverband.rb
danmayer: yep, a ENV var to run the full suite with a different default store, or the like sounds like a great plan.

.travis.yml
```diff
@@ -12,7 +12,9 @@ services:
 script:
   - bundle exec rake rubocop
   - bundle exec rake
+  - COVERBAND_HASH_REDIS_STORE=t bundle exec rake
```
danmayer: love the env toggle on tests, been doing this while rolling out dual strategies at work

lib/coverband/adapters/hash_redis_store.rb
```diff
@@ -0,0 +1,164 @@
...
+    class HashRedisStore < Base
+      FILE_KEY = 'file'
+      FILE_LENGTH_KEY = 'file_length'
+      META_DATA_KEYS = [DATA_KEY, FIRST_UPDATED_KEY, LAST_UPDATED_KEY].freeze
+      ###
+      # This key isn't related to the coverband version, but to the interal format
+      # used to store data to redis. It is changed only when breaking changes to our
+      # redis format are required.
+      ###
+      REDIS_STORAGE_FORMAT_VERSION = 'coverband_3_3'
```
danmayer: should we have a different base, so one could switch between stores if they wanted... `coverband_hash_3_3` or make it a bit more clear that you don't want to end up on the wrong store.

lib/coverband/adapters/hash_redis_store.rb
```diff
@@ -0,0 +1,164 @@
...
+            .each_with_object(keys: [key], args: [report_time, updated_time, file, file_hash, @ttl, data.length]) do |(coverage, index), hash|
+          if coverage
+            hash[:keys] << index
+            hash[:args] << coverage
+          end
+        end
+      end
+
+      def hash_incr_script
+        @hash_incr_script ||= @redis.script(:load, <<~LUA)
```
danmayer: perhaps load this from a file or something to make it easier to read?

lib/coverband/adapters/hash_redis_store.rb
```diff
@@ -0,0 +1,164 @@
...
+      ###
+      REDIS_STORAGE_FORMAT_VERSION = 'coverband_3_3'
+
+      attr_reader :redis_namespace
+
+      def initialize(redis, opts = {})
+        super()
+        @redis_namespace = opts[:redis_namespace]
+        @format_version = REDIS_STORAGE_FORMAT_VERSION
+        @redis = redis
```
danmayer: Do we need to do a check on minimum version supported or anything else to ensure the lua script will work?

danmayer:
looks, like it dropped coverage a bit... Overall looks good, let me know if you want to do anything else before getting this merge in for some alpha testing.

kbaum:
Will look into why coverage dropped. May be because I skipped the redis_store_test.rb on the second run with hash redis store env variable.

Want to do some cleanup and optimization around calculating hash of file. Currently working on that and will let you know.

lib/coverband/adapters/hash_redis_store.rb
kbaum: Done!

lib/coverband/adapters/hash_redis_store.rb
kbaum: I could do that but I have string interpolation here. Think it's worth it?

kbaum:
@danmayer - think this is ready for another look when you're free. I merged in the changes from #308 for this.

lib/coverband/adapters/hash_redis_store.rb
danmayer: Nah, we can leave as is for now

lib/coverband/collectors/coverage.rb
```diff
@@ -57,7 +57,7 @@ def report_coverage
       rescue StandardError => e
         @logger&.error 'coverage failed to store'
         @logger&.error "Coverband Error: #{e.inspect} #{e.message}"
-        @logger&.error e.backtrace if @verbose
+        e.backtrace.each { |line| @logger&.error line } if @verbose
```
danmayer: could do ` @logger&.error e.backtrace.join(', ') if @verbose` but I guess either is fine

danmayer:
How has all the additional objects impacted memory or perf of our benchmarks...

```
Coverband::Utils::AbsoluteFileConverter
Coverband::Utils::FileHasher.hash
Coverband::Utils::RelativeFileConverter
```

Overall I think it is easier to read and better structured, but want to be careful about object allocation / etc

danmayer:
OK, I pulled this down and it works on the demo site... I think given it works and this is a big PR and a new optional feature, I say we merge this and then we can do some further benchmarking and testing while it is on the master branch... Updating some documentation, and benchmarking against some real sites seems like a good next step.

lib/coverband/collectors/coverage.rb
kbaum: Reason i like this approach is if someone is prefixing each log with something like the request id or job id, then each line of the backtrace which come in a search by that id.

kbaum:
> How has all the additional objects impacted memory or perf of our benchmarks...

I added some benchmark results of just the new file path helper classes compared to master in #308 description and it actually seems to be better on memory which doesn't make much sense to me honestly.

> I say we merge this and then we can do some further benchmarking and testing while it is on the master branch.

Agreed.

danmayer:
WOOHOO congrats amazing work

Issue #215, opened 2019-03-19
re-investigate view layer support


danmayer:
While stdlib Coverage doesn't support correct runtime line coverage of compiled view templates. I believe you can get a number of hits to an individual view file. This would still be very helpful in finding dead view files. We could likely give something similar to the Gem index view without the details layer for view files... the number of times eager loaded / executed for the full file.

If we could add any support for view files back, it would be a compelling feature.

danmayer:
This, by the way, was a real-world request for folks using the gem at my work ;)

danmayer:
My first attempt at this didn't go well... I will rethink this and perhaps give it another go

jjb:
Would definitely be interested in stats on if a file got 0 hits.

danmayer:
I believe I now have a good working model for this... but need to get the current memory issues/release under control first.

felix91gr:
Would love it if coverband could track view files usage. Not even in production: just in local development would be enough for me (not sarcastic at all: if there is a way to make it track views, I don't know it x3). Although in production would of course be awesome.

I found the gem precisely whilst looking for this functionality... will keep tabs on the thread, just in case you manage to make it work @danmayer <3

danmayer:
I have found a good solution and our other release is out, I should be able to work on this sometime in the next couple weeks.

danmayer:
You can follow along with the spike here, but view layer tracking at the file level at least should make the next release. https://github.com/danmayer/coverband/pull/306

danmayer:
The view tracker is live and probably could be considered out of alpha/beta as it has significant real-world production usage. 

Closing this

PR #306, opened 2019-07-30
add support for view layer tracking

danmayer:
This is just sharing the first spike on view file tracking for Coverband. This is currently porting my old library that did view layer tracking into Coverband. It proves that the method still works. It shows that it can add value quickly to folks already using Coverband. This addresses issue #215 

Screen Shot 2019-07-29 at 10 55 29 PM

Next, I will build this to better utilize some of the patterns in Coverband like background reporting, adapters, and web view reports.

---

Usage in `config/coverband.rb`:

```
config.track_views = true
```

example console usage:

```
> CoverbandViewTracker = Coverband::Collectors::ViewTracker.new(Coverband.configuration.store.send(:redis))
 => #<Coverband::Collectors::ViewTracker:0x00007fcea1a15028 @store=#<Redis client...

> CoverbandViewTracker.unused_views
 => ["app/views/posts/show.html.erb", "app/views/posts/new.html.erb", "app/views/layouts/mailer.html.erb"]
> CoverbandViewTracker.used_views
 => ["layouts/application", "app/views/layouts/_navbar.html.erb", "app/views/posts/_form.html.erb", "app/views/home/index.html.erb", "app/views/layouts/_sidebar.html.erb", "app/views/posts/edit.html.erb", "app/views/posts/index.html.erb"]
```


danmayer:
Obviously, a bunch more work to make this really ready, but take a look @kbaum and let me know if you see any blockers with the basic idea.

danmayer:
OK, take a look @kbaum while the html report view could be made much nicer, I believe this is ready to get merged as an optional feature in the next release.

This moved from inline reporting to reporting in the background, better usage of the adapter store concept, follows the Coverband ignore filters, by default filters out gem view files, etc... I believe all the functionality is there and now before release can just work on some polish.

You can check it out on Coverband Demo,

lib/coverband/collectors/view_tracker.rb
```diff
@@ -0,0 +1,120 @@
...
+        raise NotImplementedError, 'View Tracker requires Rails 4 or greater' unless self.class.supported_version?
+
+        @project_directory = File.expand_path(Coverband.configuration.root)
+        @ignore_patterns = Coverband.configuration.ignore
+        @store = options.fetch(:store) { Coverband.configuration.store }
+        @logger = options.fetch(:logger) { Coverband.configuration.logger }
+        @target = options.fetch(:target) { DEFAULT_TARGET }
+
+        @roots = options.fetch(:roots) { Coverband.configuration.all_root_patterns }
+        @roots = @roots.split(',') if @roots.is_a?(String)
```
kbaum: Feels a little funny that we have to do this in the collector class. Something we might want to clean up later.

lib/coverband/collectors/view_tracker.rb
```diff
@@ -0,0 +1,120 @@
...
+        # such as .html.erb or .js.erb
+        # http://edgeguides.rubyonrails.org/active_support_instrumentation.html#render_partial-action_view
+        ###
+        return unless (layout_file = payload[:layout])
+        return unless newly_seen_file?(layout_file)
+
+        logged_views << layout_file
+        views_to_record << layout_file if track_file?(layout_file)
+      end
+
```
kbaum: Does it make sense to extract a method here? Something like:

```ruby
def track_views(_name, _start, _finish, _id, payload)
  track_view(payload[:identifier])
  track_view(payload[:layout])
end

private

def track_view(file)
  if file && newly_seen_file?(file)
    logged_views << file
    views_to_record << file if track_file?(file)
  end
end
```

Not sure if that'll work.

danmayer: it does work, but this is the tight loop that runs during Rails web requests, multiple times, it runs on every render call. So we want to keep it as fast as possible. For that reason, I have avoided all extra method calls. I could document that perhaps?

lib/coverband/collectors/view_tracker.rb
danmayer: Yeah, I guess right now this class combines, collector, storage, and reporting concerns. It is nice in that it is small and contains pretty much everything, but does break some of the modularize concerns we had built. 

I think for now it is OK because it is nice to keep it small and compact, and I am not sure as we look at supporting a few different plugin / tools, if the current way we split things up will be flexible enough to support  that.

lib/coverband/collectors/view_tracker.rb
kbaum: Interesting. Kind of surprised that a few method calls would make a difference. We can test this kind of thing out once we have some solid benchmarking.

danmayer: yeah, I haven't really benchmarked this, but I know when I built it a few years ago it went through a few perf iterations. I can work on getting more benchmarks... I do believe for a rails app with a partial in a loop say table row partials, this can end up with hundreds of calls in a single page render.

jjb:
Should this section of the readme be updated? https://github.com/danmayer/coverband#known-issues

danmayer:
yeah I can update that @jjb although we still don't have a line for line coverage in view files, which in some sense is what I would really like ;)

updated

lib/coverband/collectors/view_tracker.rb
```diff
@@ -0,0 +1,124 @@
...
+        @ignore_patterns = Coverband.configuration.ignore
+        @store = options.fetch(:store) { Coverband.configuration.store }
+        @logger = options.fetch(:logger) { Coverband.configuration.logger }
+        @target = options.fetch(:target) { DEFAULT_TARGET }
+
+        @roots = options.fetch(:roots) { Coverband.configuration.all_root_patterns }
+        @roots = @roots.split(',') if @roots.is_a?(String)
+
+        @logged_views = []
+        @views_to_record = []
```
hosamaly, line 33: I beg your pardon, but may I ask why an array is a better choice here than a `Set`?

danmayer: Yeah, a set would make more sense and would be faster. That is a good idea, thanks @hosamaly, feel free to send a PR or I can add it to my feature list backlog.

hosamaly: I opened PR #418

PR #311, opened 2019-08-07
Add benchmark subset of files

kbaum:
- Added benchmark of storing subset of files
- Added nil's to fake coverage data
- Fixed incorrect format for fake coverage data

### RedisStore
```
➜  coverband git:(benchmark_subset_of_files) bundle exec rake benchmarks:redis_reporting
D, [2019-08-07T09:45:48.251145 #22282] DEBUG -- : using default configuration
   store_reports_all     2.000  i/100ms
   store_reports_all     24.709  (± 4.0%) i/s -    370.000  in  15.008521s
store_reports_subset     4.000  i/100ms
store_reports_subset     40.661  (± 4.9%) i/s -    812.000  in  20.004844s
```

### HashRedisStore

```
   store_reports_all      4.535  (± 0.0%) i/s -     68.000  in  15.030560s
store_reports_subset    11.000  i/100ms
store_reports_subset     96.209  (± 7.3%) i/s -      1.925k in  20.102280s
```

### DIFF
--- a/test/benchmarks/benchmark.rake
+++ b/test/benchmarks/benchmark.rake
@@ -141,7 +141,7 @@ namespace :benchmarks do
   end
 
   def fake_line_numbers
-    24.times.each_with_object({}) do |line, line_hash|
+    28.times.each_with_object({}) do |line, line_hash|
       line_hash[(line + 1).to_s] = rand(5)
     end
   end
@@ -183,7 +183,13 @@ namespace :benchmarks do
     5.times { store.save_report(report) }
     Benchmark.ips do |x|
       x.config(time: 15, warmup: 5)
-      x.report('store_reports') { store.save_report(report) }
+      x.report('store_reports_all') { store.save_report(report) }
+    end
+    keys_subset = report.keys.first(100)
+    report_subset = report.select { |key, _value| keys_subset.include?(key) }
+    Benchmark.ips do |x|
+      x.config(time: 20, warmup: 5)
+      x.report('store_reports_subset') { store.save_report(report_subset) }
     end
   end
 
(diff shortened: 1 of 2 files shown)


test/benchmarks/benchmark.rake
```diff
@@ -140,14 +140,16 @@ namespace :benchmarks do
     Coverband::Collectors::Coverage.instance.reset_instance
   end
 
+  LINES = 45
```
kbaum: @danmayer - went into production and calculated number of lines per file and number of lines with nil. Used this script but could use a second eye on it as the number of nils was surprising to me and I wouldn't put it past myself to make a silly math mistake:

```ruby
data = Coverband.configuration.store.coverage.values.map{|hash| hash['data']}; 1
average_lines = data.map(&:length).sum/data.length
average_nil = data.map{ |file| file.count(&:nil?) }.sum/data.length
```

test/benchmarks/benchmark.rake
```diff
@@ -140,9 +140,11 @@ namespace :benchmarks do
...
+  NON_NIL_LINES = 18
   def fake_line_numbers
-    24.times.each_with_object({}) do |line, line_hash|
-      line_hash[(line + 1).to_s] = rand(5)
+    LINES.times.map do |line|
+      coverage = (line < NON_NIL_LINES) ? rand(5) : nil
```
kbaum: This was a hash but should have been an array. Somehow it was working before but started blowing up on HashRedisStore when i added nil's to the coverage data.

test/benchmarks/benchmark.rake
danmayer: hmm seems a bit high to me as well... I am wondering if we can't use avg across files as outliers perhaps can throw it off? what about total_lines / total_nils?

either way seems to point to having sparse storage and not storing nils

kbaum: Yeah.. maybe you can take a look at some of your projects and see what you come back with?

We are already not storing nils for `HashRedisStore`.

PR #315, opened 2019-09-11
add support for timestamps on view tracking

danmayer:
and for clearing all view tracking or individual files.

Screen Shot 2019-09-11 at 12 15 20 PM


kbaum:
Looks great!

PR #318, opened 2019-09-13
add better callout on unused files, in standard directories

danmayer:
makes it more clear that a file in standard directories was never loaded or used.

Screen Shot 2019-09-13 at 5 10 04 PM

Screen Shot 2019-09-13 at 3 51 32 PM

### DIFF
--- a/lib/coverband/utils/result.rb
+++ b/lib/coverband/utils/result.rb
@@ -55,11 +55,16 @@ module Coverband
       # the line-by-line coverage to zero (if relevant) or nil (comments / whitespace etc).
       def self.add_not_loaded_files(result, tracked_files)
         if tracked_files
+          # TODO: Can we get rid of this dup it wastes memory
           result = result.dup
           Dir[tracked_files].each do |file|
             absolute = File.expand_path(file)
 
-            result[absolute] ||= Coverband::Utils::LinesClassifier.new.classify(File.foreach(absolute))
+            puts "hit #{absolute} "*60 if result[absolute].nil?
+            result[absolute] ||= {
+              'data' => Array.new(File.foreach(absolute).to_a.length) { 0 },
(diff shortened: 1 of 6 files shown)


lib/coverband/utils/result.rb
```diff
@@ -55,11 +55,16 @@ def created_at
       # the line-by-line coverage to zero (if relevant) or nil (comments / whitespace etc).
       def self.add_not_loaded_files(result, tracked_files)
         if tracked_files
+          # TODO: Can we get rid of this dup it wastes memory
           result = result.dup
           Dir[tracked_files].each do |file|
             absolute = File.expand_path(file)
 
-            result[absolute] ||= Coverband::Utils::LinesClassifier.new.classify(File.foreach(absolute))
+            puts "hit #{absolute} "*60 if result[absolute].nil?
```
kbaum: Remove puts?

lib/coverband/utils/result.rb
```diff
@@ -55,11 +55,16 @@ def created_at
...
+            result[absolute] ||= {
+              'data' => Array.new(File.foreach(absolute).to_a.length) { 0 },
```
kbaum: Got interested in File#foreach. Looks like we can do a count on the enumerator instead of converting to array.

```ruby
Array.new(File.foreach(absolute).count)
```

kbaum:
Thinking it might make sense to add a test for this one? Maybe a capybara style test on the report?

lib/coverband/utils/result.rb
danmayer: thanks fixed

lib/coverband/utils/result.rb
danmayer: nice good call

PR #328, opened 2019-10-04
documentation for 2 config settings

jjb:


danmayer:
Thanks updated docs are always great!

PR #331, opened 2019-10-06
Batch report calls in hash_redis_store

kbaum:
Little bump up in cpu but much faster:

### Before:

image

```
➜  coverband git:(change_to_hash) COVERBAND_HASH_REDIS_STORE=t bundle exec rake benchmarks:redis_reporting
/Users/karlbaum/workspace/coverband/lib/coverband.rb:35: warning: already initialized constant Coverband::Adapters::RedisStore
/Users/karlbaum/workspace/coverband/lib/coverband/adapters/redis_store.rb:8: warning: previous definition of RedisStore was here
D, [2019-10-06T17:51:00.626595 #70722] DEBUG -- : using default configuration
   store_reports_all      2.029  (± 0.0%) i/s -     31.000  in  15.384839s
store_reports_subset     5.000  i/100ms
store_reports_subset     58.196  (± 8.6%) i/s -      1.155k in  20.025083s
```

### After:

i mage

```
➜  coverband git:(batch_call_lua) COVERBAND_HASH_REDIS_STORE=t bundle exec rake benchmarks:redis_reporting
   store_reports_all      4.660  (± 0.0%) i/s -     70.000  in  15.066954s
store_reports_subset    13.000  i/100ms
store_reports_subset    131.136  (± 6.9%) i/s -      2.613k in  20.049154s
```

Looks to be about 2X as fast according to our performance benchmark task.

### DIFF
--- a/lib/coverband/adapters/hash_redis_store.rb
+++ b/lib/coverband/adapters/hash_redis_store.rb
@@ -71,10 +71,12 @@ module Coverband
             updated_time: updated_time
           )
         end.to_json
+        return unless keys.any?
+
         arguments_key = [@redis_namespace, SecureRandom.uuid].compact.join('.')
         @redis.set(arguments_key, json)
         @redis.evalsha(script_id, [arguments_key])
-        @redis.sadd(files_key, keys) if keys.any?
+        @redis.sadd(files_key, keys)
       end
 
       def coverage(local_type = nil)
--- a/lib/coverband/adapters/hash_redis_store.rb
+++ b/lib/coverband/adapters/hash_redis_store.rb
@@ -56,26 +56,25 @@ module Coverband
         report_time = Time.now.to_i
         updated_time = type == Coverband::EAGER_TYPE ? nil : report_time
         script_id = hash_incr_script
-        @redis.pipelined do
-          keys = report.map do |file, data|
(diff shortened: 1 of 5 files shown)


lib/coverband/adapters/hash_redis_store.rb
```diff
@@ -56,26 +56,25 @@ def save_report(report)
...
+          script_input(
+            key: key,
+            file: relative_file,
+            file_hash: file_hash,
+            data: data,
+            report_time: report_time,
+            updated_time: updated_time
+          )
+        end.to_json
+        arguments_key = [@redis_namespace, SecureRandom.uuid].compact.join('.')
```
danmayer: do these other steps need to be wrapped in `if keys.any?`

kbaum: Good question. I don't see the harm in it and it might save us unnecessary hits to redis. Good thing is we do have a test for it so think its currently functional.

https://github.com/danmayer/coverband/blob/master/test/coverband/adapters/hash_redis_store_test.rb#L31

kbaum:
@danmayer  - to be clear, the cpu usage on redis still goes down about 20-25% compared to master for our benchmarks. I was comparing this iteration with the change_to_hash branch.

PR #334, opened 2019-10-12
Use flushdb on a test redis db

kbaum:
Been struggling with some of the tests sporadically failing. 

I like this approach for cleaning redis as we easily start with an empty clean each time but we avoid clearing out the data in the default redis database.

### DIFF
--- a/test/coverband/reporters/base_test.rb
+++ b/test/coverband/reporters/base_test.rb
@@ -3,6 +3,10 @@
 require File.expand_path('../../test_helper', File.dirname(__FILE__))
 
 class ReportsBaseTest < Minitest::Test
+  def setup
+    super
+  end
+
   test '#merge_arrays basic merge preserves order and counts' do
     first = [0, 0, 1, 0, 1]
     second = [nil, 0, 1, 0, 0]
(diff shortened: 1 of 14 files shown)


test/coverband/reporters/base_test.rb
```diff
@@ -47,45 +46,4 @@ class ReportsBaseTest < Minitest::Test
...
     assert_equal expected, Coverband::Reporters::Base.send(:get_current_scov_data_imp, store, roots)[:merged]
   end
-
-  ###
-  # This test uses real world example data which helped uncover a bug
-  # The copied data doesn't format easily and isn't worth the effort to meet
-  # string style
-  # rubocop:disable all
-  ###
-  test '#get_current_scov_data_imp merges multiples of file data' do
```
kbaum: This test was failing with missing mock expectations. I struggled with it for a while and thought maybe it wasn't worth maintaining. Let me know what you think.

danmayer: I guess if this blocks the PR comment out and I will take a look and see if I can quickly fix or I will remove

kbaum: Ok sounds good.

test/test_helper.rb
```diff
@@ -30,20 +30,24 @@
 
 module Coverband
   module Test
+    TEST_DB = 2
+
+    def self.redis
```
kbaum: @danmayer - added a test redis instance that all of our tests can use. This way we all our in the same db and our database is being cleaned out properly.

PR #343, opened 2019-10-18
avoid running coverband on known tasks, capture errors and log warnin…

danmayer:
…gs on bootup, avoid connecting to Redis unless it is actually used...

---

OK, this resolves some of the recent issues where startup on various jobs that don't need coverage would fail. This takes a couple of approaches, due to we have slightly different issues between the normal adapter the new HashRedisAdapter and with the view tracker.

1. This expands the rake task ignore list to include a bunch of common tasks we would likely want to ignore. This should help folks not have to do anything special when running things like `asset:precompile` on heroku or in CI environments (that have yet to stand up Redis).
2. This rescues Redis connection errors a few different places where they can occur (railtie, startup configuration, etc). This will output a warning in cases where the task isn't covered by our default ignore, but the user wouldn't like Coverband to block their app startup
3. For the view tracker, move from hitting Redis on initialization to lazily hit during runtime. This avoids startup costs when the view tracker will never actually be used (in rake tasks for example). @bhb I believe this is why you started to get the error on the upgrade. The old view_tracker didn't hit Redis in the initialization, and the next version did, which would have surfaced this error. 
    a. Note: The old Redis adapter doesn't connect eagerly either, but the new adapter does to ensure the Redis server is new enough to support LUA scripting, if we could come up with a good way to avoid that during initialization that could be good, but I don't think we want to defer the check until runtime only as it could be confusing, I don't have a good idea for that yet.

I ran with old adapter, new, and with view_tracker (on and off), with Redis turned off and this either works silently or outputs log warnings and moves on with the task as one would expect.

### DIFF
--- a/lib/coverband/configuration.rb
+++ b/lib/coverband/configuration.rb
@@ -32,8 +32,7 @@ module Coverband
                     'db:setup',
                     'db:structure:dump',
                     'db:structure:load',
-                    'db:version'
-                  ]
+                    'db:version']
 
     # Heroku when building assets runs code from a dynamic directory
     # /tmp was added to avoid coverage from /tmp/build directories during
--- a/lib/coverband/configuration.rb
+++ b/lib/coverband/configuration.rb
@@ -23,7 +23,17 @@ module Coverband
     IGNORE_TASKS = ['coverband:clear',
                     'coverband:coverage',
                     'coverband:coverage_server',
-                    'coverband:migrate']
+                    'coverband:migrate',
+                    'assets:precompile',
+                    'db:version',
+                    'db:create',
+                    'db:drop',
+                    'db:seed',
+                    'db:setup',
+                    'db:structure:dump',
+                    'db:structure:load',
+                    'db:version'
+                  ]
 
     # Heroku when building assets runs code from a dynamic directory
     # /tmp was added to avoid coverage from /tmp/build directories during
--- a/lib/coverband/utils/railtie.rb
+++ b/lib/coverband/utils/railtie.rb
@@ -29,9 +29,11 @@ module Coverband
     end
 
     config.after_initialize do
-      Coverband.eager_loading_coverage!
-      Coverband.report_coverage
-      Coverband.runtime_coverage!
+      unless Coverband.tasks_to_ignore?
+        Coverband.eager_loading_coverage!
+        Coverband.report_coverage
+        Coverband.runtime_coverage!
+      end
     end
 
     rake_tasks do
--- a/lib/coverband/utils/railtie.rb
+++ b/lib/coverband/utils/railtie.rb
@@ -11,15 +11,20 @@ module Coverband
 
   class Railtie < Rails::Railtie
     initializer 'coverband.configure' do |app|
-      app.middleware.use Coverband::BackgroundMiddleware
+      begin
+        app.middleware.use Coverband::BackgroundMiddleware
 
-      if Coverband.configuration.track_views
-        CoverbandViewTracker = Coverband::Collectors::ViewTracker.new
-        Coverband.configuration.view_tracker = CoverbandViewTracker
+        if Coverband.configuration.track_views
+          CoverbandViewTracker = Coverband::Collectors::ViewTracker.new
+          Coverband.configuration.view_tracker = CoverbandViewTracker
 
-        ActiveSupport::Notifications.subscribe(/render_partial.action_view|render_template.action_view/) do |name, start, finish, id, payload|
-          CoverbandViewTracker.track_views(name, start, finish, id, payload) unless name.include?('!')
+          ActiveSupport::Notifications.subscribe(/render_partial.action_view|render_template.action_view/) do |name, start, finish, id, payload|
+            CoverbandViewTracker.track_views(name, start, finish, id, payload) unless name.include?('!')
+          end
         end
+      rescue Redis::CannotConnectError => error
+        Coverband.configuration.logger.info "Redis is not available (#{error}), Coverband not configured"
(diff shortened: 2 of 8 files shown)


lib/coverband/utils/railtie.rb
```diff
@@ -11,22 +11,29 @@ def eager_load!
...
+          CoverbandViewTracker = Coverband::Collectors::ViewTracker.new
+          Coverband.configuration.view_tracker = CoverbandViewTracker
 
-        ActiveSupport::Notifications.subscribe(/render_partial.action_view|render_template.action_view/) do |name, start, finish, id, payload|
-          CoverbandViewTracker.track_views(name, start, finish, id, payload) unless name.include?('!')
+          ActiveSupport::Notifications.subscribe(/render_partial.action_view|render_template.action_view/) do |name, start, finish, id, payload|
+            CoverbandViewTracker.track_views(name, start, finish, id, payload) unless name.include?('!')
+          end
         end
+      rescue Redis::CannotConnectError => error
```
kbaum: Would this not be covered by the others rescue block. Think we could reproduce redis being not available with a forked test pointing at a redis instance that does not exist.

danmayer: railtie exceptions can't be covered by the outer block on setup here, because you can't rescue around the `require`, the error doesn't propagate to where you require. 

https://github.com/danmayer/coverband/pull/343/files#diff-a637af9972be5b34783d41387d25e0bfR110

I will see if I can get a test around it

kbaum: Oh gotcha. Makes sense.

danmayer:
OK, I got a test added that confirms we don't end up failing on rake tests when Coverband can't connect to the redis @kbaum

lib/coverband/utils/railtie.rb
kbaum: Just thinking about it some more.. this one may be challenging to test since this code executes as soon as coverband is required.

test/forked/rails_rake_full_stack_test.rb
```diff
@@ -45,4 +45,10 @@ def setup
     assert_equal empty_hash, coverage_report[:eager_loading]
     assert_equal empty_hash, coverage_report[:merged]
   end
+
+  test "doesn't exit non-zero with error on missing redis" do
+    output = `COVERBAND_CONFIG=./test/rails#{Rails::VERSION::MAJOR}_dummy/config/coverband.rb bundle exec rake -f test/rails#{Rails::VERSION::MAJOR}_dummy/Rakefile coverband:clear`
```
kbaum: Great idea!

lib/coverband/configuration.rb
```diff
@@ -23,7 +23,16 @@ class Configuration
     IGNORE_TASKS = ['coverband:clear',
                     'coverband:coverage',
                     'coverband:coverage_server',
-                    'coverband:migrate']
+                    'coverband:migrate',
+                    'assets:precompile',
+                    'db:version',
```
gingerlime: what about whitelisting all `db:.*` commands? e.g. `db:migrate` in particular jumps to mind, maybe also `db:test:prepare`, and I'm sure there are others... but I wonder if this is intentional to run coverage on migrations or other db tasks perhaps?

side note: https://github.com/danmayer/coverband/blob/7ce04ee217f750edd413e4a371c474c933615fe6/lib/coverband.rb#L76 checks for `tasks_to_ignore?` twice on two lines... maybe a guard clause will be a bit tidier and more efficient? .... aaaand, wouldn't it be more sensible to guard at the beginning of the `start` method, https://github.com/danmayer/coverband/blob/e7917fc7c247877db863f130bae2c7784c8d5d72/lib/coverband.rb#L73 so you don't even need to instantiate anything?

One thing I noticed is that I see redis errors, even on commands that I'd expect coverband to skip, so perhaps this is related??

gingerlime: I obviously can only speak for myself, but find it rather hard to imagine people care much about coverage reports for their migrations or other db related tasks... But it's definitely possible...

The practical distinction between `tasks_to_ignore` and `tasks_to_exclude_coverage` seems rather elusive to me. `ignore` implies not doing anything at all (to me), but what's the difference if you're "just" excluding coverage? I'm not sure... Definitely adds more complexity and I can't imagine it's worth it.

> but we still need to configure Coverband and instantiate it so that we can build the report, or clear or whatever the task was doing

I'm probably missing something here, but when would you need Coverband instantiated but not reporting coverage?

Completely agree that `assets:precompile` is a very likely candidate to ignore. On Heroku, Docker builds, and probably in other cases.

Please forgive me if I'm over-simplifying things because of ignorance or lack of familiarity with different use-cases. Personally however it feels to me that the simplest option is usually best. Either do nothing, or do everything... 

In any case, it's definitely an improvement already, but just wanted to contribute my 2 cents.

Thanks for fixing this so quickly and taking the time to share your reasoning and discuss this.

kbaum:
I think we should try to ignore as little as possible. Adding special logic to disable coverage collection makes the library more difficult to reason about which reduces confidence in the generated reports.

gingerlime:
Having given it a bit more thought, I tend to agree with @kbaum. Keep it simple and enabled by default (but fail safe!). If someone doesn't want coverage in rake tasks etc, they can exclude it using the ENV or set up their own ignore list.

Less magic. Less surprises. Simpler code. No arguments over what should/not be excluded :)

Just an outsider's perspective. I know it's a tricky choice.

kbaum:
> Less magic. Less surprises. Simpler code. No arguments over what should/not be excluded :)

Well put!

PR #346, opened 2019-10-23
Fix ignore issues

danmayer:
These two issues came up while configuring Coverband for a new app... The vendor one, in particular, makes it look like we are missing data.

The move from include to match, is to keep the filters in sync we were allowing files in that we would filter out during reporting... This can lead to a lot of extra work for storage and merging.

### DIFF
--- a/test/coverband/collectors/coverage_test.rb
+++ b/test/coverband/collectors/coverage_test.rb
@@ -82,6 +82,15 @@ class CollectorsCoverageTest < Minitest::Test
     assert_equal false, @coverband.send(:track_file?, heroku_build_file)
   end
 
+  # verifies a fix where we were storing, merging, and tracking ignored files
+  # then just filtering them out of the final report
+  test 'ignores uses regex same as reporter does' do
+    regex_file = Coverband.configuration.current_root + '/config/initializers/fake.rb'
+    assert_equal true, @coverband.send(:track_file?, regex_file)
+    @coverband.instance_variable_set(:@ignore_patterns, ['config/initializers/*'])
+    assert_equal false, @coverband.send(:track_file?, regex_file)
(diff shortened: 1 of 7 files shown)


test/coverband/collectors/coverage_test.rb
```diff
@@ -82,6 +82,15 @@ def teardown
     assert_equal false, @coverband.send(:track_file?, heroku_build_file)
   end
 
+  # verifies a fix where we were storing, merging, and tracking ignored files
+  # then just filtering them out of the final report
+  test 'ignores uses regex same as reporter does' do
+    regex_file = Coverband.configuration.current_root + '/config/initializers/fake.rb'
+    assert_equal true, @coverband.send(:track_file?, regex_file)
```
kbaum: I think if possible we should try and avoid calling private methods and instance vars from our tests. Can make refactoring difficult later.

danmayer: I know there is some advice about not testing private methods, I think we discussed it in another previous PR. I disagree with that testing advice as it makes tests far more complicated and makes it harder to interpret results. I think we will have different styles on this level of testing, which is cool as any project that has more than one dev has some divergent patterns.

Being able to target a highly specific method where complexity resides has at least for me, shown good value vs the tradeoffs of having to refactor tests when they change.

kbaum: Makes sense. Apologize for bringing the same issue again. Sometimes I think needing to test a private method, might be a good reason to extract the complicated logic into a class of its own.

danmayer: ha no worries, this seems to be one of the few things we have different coding opinions on ;)

Issue #319, opened 2019-09-20
Stale file data being reported when new file untouched


kbaum:
If a file has had no coverage but then has a change made to the file which updates the checksum, the report will show the coverage data for the old version of the file with the new version of the source code. This most often is not an issue as the file is usually covered and when the data is sent to redis, the coverage on the new version of the file with the new checksum replaces the old coverage data.

First reported in #312

danmayer:
fixed closing this one out

PR #348, opened 2019-10-23
fixes that we shouldn't report coverage if all the data is out of sync

danmayer:
This is a fix for #319 which came up a while ago, wanted to get the fix in before the next release.

### DIFF
--- a/test/coverband/adapters/redis_store_test.rb
+++ b/test/coverband/adapters/redis_store_test.rb
@@ -36,6 +36,15 @@ unless ENV['COVERBAND_HASH_REDIS_STORE']
       assert current_time <= @store.coverage['app_path/dog.rb']['last_updated_at']
     end
 
+    def test_file_hash_change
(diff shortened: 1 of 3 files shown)


test/coverband/adapters/redis_store_test.rb
```diff
@@ -36,6 +36,15 @@ def test_coverage_increments
       assert current_time <= @store.coverage['app_path/dog.rb']['last_updated_at']
     end
 
+    def test_file_hash_change
+      mock_file_hash(hash: 'abc')
+      @store.save_report('app_path/dog.rb' => [0, nil, 1, 2])
+      assert_equal [0, nil, 1, 2], @store.coverage['app_path/dog.rb']['data']
+      mock_file_hash(hash: '123')
+      $debug = true
```
kbaum: Still need this?

danmayer: oops no, I will remove

PR #354, opened 2019-11-02
Use pipeline in call to coverage in hash_redis_store

kbaum:
Benchmark comparison of connecting to rediscloud instance with large dataset remotely from local machine:

```ruby
puts Benchmark.measure { Coverband.configuration.store.coverage  }
```

Before (master)
```
5.117984   0.857738   5.975722 (108.416087)
```

This branch
```
2.947418   0.617277   3.564695 (  7.217492)
```


kbaum:
Let’s definitely do a release candidate.

PR #357, opened 2019-11-07
Use msgpack

kbaum:
### Before

image

```
   store_reports_all      3.192  (± 0.0%) i/s -     48.000  in  15.146428s
store_reports_subset     9.000  i/100ms
store_reports_subset     95.632  (± 9.4%) i/s -      1.899k in  20.066744
```

### After

image

```
   store_reports_all      3.564  (± 0.0%) i/s -     54.000  in  15.242361s
store_reports_subset    10.000  i/100ms
store_reports_subset    102.395  (± 9.8%) i/s -      2.030k in  20.031830s
```

Using msg_pack gives around a 6.8% speed increase along with a moderate decrease in memory usage.  CPU utilization does go up but that seems to correlate to more commands per second.

### DIFF
--- a/coverband.gemspec
+++ b/coverband.gemspec
@@ -22,6 +22,7 @@ Gem::Specification.new do |spec|
   # to test support for sdk 1, uncomment this line
   # spec.add_development_dependency 'aws-sdk', '~> 1'
   # to test sdk 2 use this one
+  spec.add_dependency 'msgpack'
   spec.add_development_dependency 'aws-sdk-s3', '~> 1'
   spec.add_development_dependency 'benchmark-ips'
   spec.add_development_dependency 'capybara'
--- a/lib/coverband.rb
+++ b/lib/coverband.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'logger'
+require 'msgpack'
 require 'json'
 require 'redis'
 require 'coverband/version'
--- a/lib/coverband/adapters/hash_redis_store.rb
+++ b/lib/coverband/adapters/hash_redis_store.rb
@@ -77,7 +77,7 @@ module Coverband
           next unless files_data.any?
 
           arguments_key = [@redis_namespace, SecureRandom.uuid].compact.join('.')
-          @redis.set(arguments_key, { ttl: @ttl, files_data: files_data }.to_json, ex: JSON_PAYLOAD_EXPIRATION)
+          @redis.set(arguments_key, { ttl: @ttl, files_data: files_data }.to_msgpack, ex: JSON_PAYLOAD_EXPIRATION)
           @redis.evalsha(hash_incr_script, [arguments_key])
         end
         @redis.sadd(files_key, keys) if keys.any?
--- a/lua/install.sh
+++ b/lua/install.sh
@@ -13,6 +13,6 @@ pip install hererocks
 hererocks $LUA_DIR -l5.1 -rlatest
 source $LUA_DIR/bin/activate
 lua -v
-for i in luacov busted redis-lua inspect lua-cjson; do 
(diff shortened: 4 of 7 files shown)


coverband.gemspec
```diff
@@ -22,6 +22,7 @@ Gem::Specification.new do |spec|
   # to test support for sdk 1, uncomment this line
   # spec.add_development_dependency 'aws-sdk', '~> 1'
   # to test sdk 2 use this one
+  spec.add_dependency 'msgpack'
```
danmayer: I would try to make this an optional dependency... We want to be really careful about adding any additional dependencies... I guess with JSON built-in and only 6% better not sure if it is worth it... If so definitely feels more like an opt-in feature vs the default.

Thoughts?

kbaum: Agreed but I want to see how msg_pack compares when we increase the batch size of save_report.

danmayer: yeah I think you could just not have it as a dependency and then have something like a flag that would use it, and in that case it would require the user to explicitly add msgpack to their gemfile...

All that being said for testing it is obviously totally fine.

kbaum: Makes sense.

kbaum:
With a 500 batch size, there is a 15.5% difference in save_report and significant less memory used by msg_pack... maybe 40% looking at the graph?

### Message pack
image


```
   store_reports_all      5.687  (± 0.0%) i/s -     86.000  in  15.142712s
store_reports_subset    16.000  i/100ms
store_reports_subset    165.940  (± 2.4%) i/s -      3.328k in  20.068828s
```

### JSON

image

```
   store_reports_all      4.821  (± 0.0%) i/s -     73.000  in  15.176836s
store_reports_subset    14.000  i/100ms
store_reports_subset    144.801  (± 6.2%) i/s -      2.884k in  20.002879s
```

danmayer:
this is on the Redis CPU / memory side right? Have you run any of our benchmarks or memory tests to see the impact on the Ruby side? I am really curious if msgpack would reduce the memory load on the server as well as I believe it does far better 'nil' handling than `to_json` which creates the string `null` in memory over and over.

kbaum:
Ahhh.. right. This is on the redis side. I am using [redis-stat](https://github.com/junegunn/redis-stat). Great point about memory on ruby. Will do some benchmarks and report back..

kbaum:
# Ruby Memory Benchmarks

Tested with:
```
bundle exec rake benchmarks:run_big
```

For HashRedisStore there are very minor memory savings in ruby when using msg_pack. Interesting to note though that HashRedisStore allocates much less memory than the default RedisStore. I haven't yet modified RedisStore to use msg_pack but that would be interesting to measure.

## HashRedisStore

### JSON

```
Total allocated: 213725850 bytes (1139205 objects)
Total retained:  664216 bytes (4015 objects)

allocated memory by gem
-----------------------------------
 209819690  coverband/lib
   3354010  redis-4.1.2
    529870  other
     22280  securerandom
```


### msgpack

```
Total allocated: 207166250 bytes (995245 objects)
Total retained:  664216 bytes (4015 objects)

allocated memory by gem
-----------------------------------
 201464010  coverband/lib
   3117730  redis-4.1.2
   2032360  msgpack-1.3.1
    529870  other
     22280  securerandom
```

## RedisStore

```
D, [2019-11-09T05:17:21.315395 #67377] DEBUG -- : using default configuration
Total allocated: 282642545 bytes (1544223 objects)
Total retained:  664216 bytes (4015 objects)

allocated memory by gem
-----------------------------------
 222782660  coverband/lib
  30916130  json
  19585080  redis-4.1.2
   9358675  other

allocated memory by file
```

PR #364, opened 2020-01-10
fix typo

maxp-hover:
It would be useful if these "search paths" were configurable by the user as well. However I am not taking that on right now.


danmayer:
nice, thanks this is great @maxp-hover

PR #369, opened 2020-03-26
Match index route without trailing /

rswaminathan:
In my rails 4,`request.path_info` is an empty string instead of "/". This fix addresses that.


danmayer:
Thanks @rswaminathan I will test this on a few apps and try to get it in soon. I am also going to see if I can fix without overwriting the `request.path`

danmayer:
ok @rswaminathan  you should be able to install the release candidate `4.2.5.rc.1` using that version in your gemfile to verify the fix works for you.

Issue #368, opened 2020-02-06
Error when using simplecov 18.1


hirowatari:
I upgraded from simplecov 17.1 to 18.1.  Now when I run tests I get the error

```
E, [2020-02-06T15:41:28.180442 #1646] ERROR -- : coverage failed to store
E, [2020-02-06T15:41:28.180546 #1646] ERROR -- : Coverband Error: #<NoMethodError: undefined method `nonzero?' for [:lines, [nil, nil, 1, 1, 1, 1, nil, nil]]:Array> undefined method `nonzero?' for [:lines, [nil, nil, 1, 1, 1, 1, nil, nil]]:Array
```

Is this a coverband issue or simplecov?  Does coverband conflict with the latest simplecov?

danmayer:
did you enable branch support for simplecov? If so that will change what stdlib `Coverage` reports so that coverband can't process it. 

I can look at being able to handle both formats, but for now I would say SimpleCov is for test env and Coverband is for dev,staging,production... So another easy fix would be to put `, require: false` after coverband in your Gemfile. Then in your your `application.rb` you can manually start it in non test environments.

```
Bundler.require(*Rails.groups)
#add after bundler require
require 'coverband' unless Rails.env.test?
```

We have support for oneshot while simplcov does not, so the same sort of error would occur if you enabled oneshot in coverband, simplecov wouldn't be able to parse the output of Coverage... I am sure at some point both tools will be able to process all 3 Coverage formats ;)

hirowatari:
Thanks for the quick response. No I didn't enable branch coverage.

Your solution worked.  It makes sense to not run coverband in test.  Thanks.

hirowatari:
One small addition is that you'll need to edit your routes to avoid referencing coverband in test eg:
`mount Coverband::Reporters::Web.new, at: '/coverage' if defined? Coverband`

danmayer:
thanks for noticing the route issue and adding the comment. I am thinking we can detect this case and log out a warning and then safely ignore it. It seems like a good fix until we add full support.  I am guessing SimpleCov must have enabled by default.

danmayer:
this PR handles the new formats and issues with both Coverband and SimpleCov in the same project https://github.com/danmayer/coverband/pull/371

PR #371, opened 2020-04-12
handle stdlib branch coverage options, now a default in simplecov 0.1…

danmayer:
…8.x we will get more bug reports for projects that use both

This handles two issues:

* we need to be able to parse the new formats in stdlib Ruby Coverage 2.5 & 2.6, namely :lines which comes along with branching support, since SimpleCov 0.18.x the :lines format is a default
* also, Simplecov doesn't check if Coverage is already running, and we start Coverage first which causes an error if both are configured. I detect that case and output a warning, as Coverage will be started late and miss some files our eager_loading hook

The first fix is really valuable as anyone can configure the various Coverage format options, the simplecov specific detecting is a bit annoying and specific to if both libs are trying to run in test mode. For now, we don't really recommend running Coverband in test.

this was tested on Coverband Demo with SimpleCov 0.18 setup, resolves https://github.com/danmayer/coverband/issues/368

### DIFF
--- a/lib/coverband/collectors/coverage.rb
+++ b/lib/coverband/collectors/coverage.rb
@@ -89,12 +89,17 @@ module Coverband
         raise NotImplementedError, 'Coverage needs Ruby > 2.3.0' if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.3.0')
 
         require 'coverage'
-        if Coverage.ruby_version_greater_than_or_equal_to?('2.6.0')
-          ::Coverage.start(oneshot_lines: Coverband.configuration.use_oneshot_lines_coverage) unless ::Coverage.running?
-        elsif Coverage.ruby_version_greater_than_or_equal_to?('2.5.0')
-          ::Coverage.start unless ::Coverage.running?
+        if defined?(SimpleCov) && defined?(Rails) && defined?(Rails.env) && Rails.env.test?
+          puts "Coverband: detected SimpleCov in test Env, allowing it to start Coverage"
+          puts "Coverband: to ensure no error logs or missing Coverage call `SimpleCov.start` prior to requiring Coverband"
         else
-          ::Coverage.start
+          if Coverage.ruby_version_greater_than_or_equal_to?('2.6.0')
+            ::Coverage.start(oneshot_lines: Coverband.configuration.use_oneshot_lines_coverage) unless ::Coverage.running?
+          elsif Coverage.ruby_version_greater_than_or_equal_to?('2.5.0')
+            ::Coverage.start unless ::Coverage.running?
+          else
+            ::Coverage.start
+          end
         end
         reset_instance
       end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -46,6 +46,10 @@ Style/ClassVars:
   Enabled: false
 Style/MultilineBlockChain:
   Enabled: false
+Style/StringLiterals:
+  Enabled: false
+Style/IfInsideElse:
+  Enabled: false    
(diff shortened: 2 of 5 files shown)


danmayer:
This is the issue by the way, https://github.com/colszowka/simplecov/blob/4e405d5fe0520a43f6757039444a423fed928c6a/lib/simplecov.rb#L347

I might submit a PR over there to not call if already running.

lib/coverband/collectors/coverage.rb
```diff
@@ -89,12 +89,17 @@ def initialize
         raise NotImplementedError, 'Coverage needs Ruby > 2.3.0' if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.3.0')
 
         require 'coverage'
-        if Coverage.ruby_version_greater_than_or_equal_to?('2.6.0')
-          ::Coverage.start(oneshot_lines: Coverband.configuration.use_oneshot_lines_coverage) unless ::Coverage.running?
-        elsif Coverage.ruby_version_greater_than_or_equal_to?('2.5.0')
-          ::Coverage.start unless ::Coverage.running?
+        if defined?(SimpleCov) && defined?(Rails) && defined?(Rails.env) && Rails.env.test?
+          puts "Coverband: detected SimpleCov in test Env, allowing it to start Coverage"
```
kbaum, line 93: What happens if `Coverage.start` is called twice? I had thought it raises an error private to 2.5 but i just tried in 2.4.4 and nothing seemed to happen.

```
➜ ruby -v
ruby 2.4.4p296 (2018-03-28 revision 63013) [x86_64-darwin17]
➜ irb
irb(main):001:0> require 'coverage'
=> true
irb(main):002:0> Coverage.start
=> nil
irb(main):003:0> Coverage.start
=> nil
irb(main):004:0>
```

danmayer: So it depends what is going on and the version of Ruby, but it raises an error for sure if you call start with different options. We call `Coverage.start` and now SimpleCov calls `Coverage.start(lines: true, branches: true)` so you get this error, when SimpleCov tries to start up.

```cannot change the measuring target during coverage measurement
/Users/danmayer/.rvm/gems/ruby-2.6.2/gems/simplecov-0.18.5/lib/simplecov.rb:354:in `start'
/Users/danmayer/.rvm/gems/ruby-2.6.2/gems/simplecov-0.18.5/lib/simplecov.rb:354:in `start_coverage_with_criteria'
/Users/danmayer/.rvm/gems/ruby-2.6.2/gems/simplecov-0.18.5/lib/simplecov.rb:343:in `start_coverage_measurement'
/Users/danmayer/.rvm/gems/ruby-2.6.2/gems/simplecov-0.18.5/lib/simplecov.rb:58:in `start'
/Users/danmayer/projects/coverband_demo/test/test_helper.rb:3:in `<top (required)>'
```

kbaum: Right. Just wondering if we could eventually simplify this part of the code by just catching the exception that gets thrown if you call start in certain ruby versions. I usually don't like to have logic that depends on exceptions being thrown but this could reduce the number of branches we have here.

danmayer: It would be hard since it is thrown from a simplecov call, and all their call stack... I guess another option is to detect SimpleCov and the version and start coverage with the same arguments that they do... in some sense I don't want to get to tied to them, and think the warning might be less complexity for now

kbaum: Gotcha. This does sound challenging.

PR #374, opened 2020-04-26
Move filters drop gem support

danmayer:
I think what is currently on master, I will release as the last 4.x.x release... I bumped version to a 5.x release candidate on this work. 

As we move toward 5.0 we are working towards simplification of the codebase:

* increased focus on performance
* reduce complexity and make it easier to maintain

Currently the changes in this branch:

* move filtering into Delta, to avoid calculating coverage diffs for files that are captured by Ruby Coverage, but are ignored
* drops Gem support as this was never performant enough for production
* drops S3 / static web reports, in favor of making it easier to support the web-ui


lib/coverband/configuration.rb
```diff
@@ -100,19 +100,19 @@ def password
...
     end
 
     def s3_access_key_id
-      @s3_access_key_id || ENV['AWS_ACCESS_KEY_ID']
+      puts 'deprecated, s3 is no longer support'
     end
 
     def s3_secret_access_key
-      @s3_secret_access_key || ENV['AWS_SECRET_ACCESS_KEY']
+      puts 'deprecated, s3 is no longer support'
```
kbaum, line 115: Good thinking not just removing so this change is backwards compatible.

PR #377, opened 2020-06-23
Address Redis `exists` deprecation warning

baffers:
After upgrading an app that has Coverband to from Redis 4.1 to [Redis 4.2.1](https://github.com/redis/redis-rb/blob/0546fe6c62de81843e8f0e9353cafba52d76eab8/CHANGELOG.md#421) I'm seeing the following deprecation warning:

> `Redis#exists(key)` will return an Integer in redis-rb 4.3. `exists?` returns a boolean, you should use it instead. To opt-in to the new behavior now you can set Redis.exists_returns_integer =  true. To disable this message and keep the current (boolean) behaviour of 'exists' you can set `Redis.exists_returns_integer = false`, but this option will be removed in 5.0.

To address this warning, this PR will updates the code to use the `exists?` method if it is defined otherwise it will continue to use the existing logic (`exists`).

### DIFF
--- a/lib/coverband/collectors/view_tracker.rb
+++ b/lib/coverband/collectors/view_tracker.rb
@@ -108,7 +108,7 @@ module Coverband
       end
 
       def report_views_tracked
-        redis_store.set(tracker_time_key, Time.now.to_i) unless @one_time_timestamp || redis_store.exists(tracker_time_key)
+        redis_store.set(tracker_time_key, Time.now.to_i) unless @one_time_timestamp || tracker_time_key_exists?
         @one_time_timestamp = true
         reported_time = Time.now.to_i
         views_to_record.each do |file|
@@ -143,6 +143,14 @@ module Coverband
 
       def redis_store
         store.raw_store
+      end
+
+      def tracker_time_key_exists?
+        if defined?(redis_store.exists?)
+          redis_store.exists?(tracker_time_key)
+        else
+          redis_store.exists(tracker_time_key)
+        end
       end
 
       def tracker_key


baffers:
hey @danmayer I was hoping to open a PR to address some deprecation warnings I'm seeing in our apps when Redis is upgrading to 4.2.*, but I realized that if an app is on Redis 4.1 or earlier, it will probably raise an exception because `exists?` isn’t a method. Tbh, I'm really not in the open source game and haven't worked on gems at all so I have no idea how to "fix" this deprecation warning so that it doesn't potentially break an app running an earlier version of redis. Any advice?

baffers:
@danmayer Ah that makes complete sense 🤦‍♀️ , thanks for your speedy feedback! I updated the code and tests are passing for me locally (although I noticed master is red).

danmayer:
ok, later tonight I will try to figure out what is up with master and to get this merged in and cut a small release. I am currently mostly working on the 5.0 branch ;)

danmayer:
OK, @baffers I forgot I had merged my 5.0 branch but decided I wanted to get a few more features in... So I could cut another 5.0 beta release that includes your change, or your change will go out when I get to 5.0... Or I could merge your commit into the old 4.5.0 and release a 4.5.1... 

Let me know how much the warnings are bothering y'all... I think you all could test out a 5.0 beta, I have a couple production systems on it... Mostly it drops S3 support which I don't think y'all use, and it adds JRuby support and refactors some internals... So I don't think it would cause you issues... Let me know your thoughts

danmayer:
Thanks @baffers since we got another report of this issue, https://github.com/danmayer/coverband/issues/379, I decided to cut a 4.2.6 release with this redis fix since the 5.0 release still needs some work...

So upgrading to 4.2.6 should resolve the issue, thanks again!

baffers:
🎉 thanks @danmayer!!

PR #388, opened 2020-08-08
Fix missing web report require statements

### LINKED ISSUE #387: uninitialized constant Coverband::Reporters::HTMLReport
image

It looks like we are trying to re-define the `Coverband::Reporters::Web.initialize`

https://github.com/danmayer/coverband/blob/78a1da323ecabc6ad780b272034854a5c3700c63/lib/coverband.rb#L116-L126

But for some reason when I debug in pry it seems the old initialize method here is being called:

https://github.com/danmayer/coverband/blob/78a1da323ecabc6ad780b272034854a5c3700c63/lib/coverband/reporters/web.rb#L14-L16

As a result, we are missing some required files.

Any reason why we redefine it in coverband.rb instead of just defining it directly here: 

https://github.com/danmayer/coverband/blob/78a1da323ecabc6ad780b272034854a5c3700c63/lib/coverband/reporters/web.rb#L14-L16

kbaum:
Fixes #387

Is this ok to just do the requires here?

### DIFF
--- a/lib/coverband/reporters/web.rb
+++ b/lib/coverband/reporters/web.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "coverband"
+
 begin
   require "rack"
 rescue LoadError
@@ -10,10 +12,6 @@ module Coverband
   module Reporters
     class Web
       attr_reader :request
-
-      def initialize
-        init_web
-      end
 
       def init_web
         full_path = Gem::Specification.find_by_name("coverband").full_gem_path


lib/coverband/reporters/web.rb
```diff
@@ -12,9 +12,18 @@ class Web
       attr_reader :request
 
       def initialize
```
danmayer, line 14: The duplicate requires here I guess is fine in theory... but how do you ever have this file loaded, the only place it was required you now have removed.

lib/coverband/reporters/web.rb
```diff
@@ -12,9 +12,18 @@ class Web
...
+        require "coverband/reporters/web"
```
danmayer, line 15: do you manually require this file somewhere in your app / initializer or whatever?

`require "coverband/reporters/web"` as I removed it from the top level gem require nothing should load it without the override.

lib/coverband/reporters/web.rb
kbaum: Oh i didn't notice the duplicate requires.

lib/coverband/reporters/web.rb
kbaum: ```"coverband/reporters/web"``` is being manually required. So this is in Gemfile:

```
gem 'coverband', require: false
```

This is in routes.rb

```ruby
require 'coverband/reporters/web'
```

And this is in application.rb

```ruby
Bundler.require(*Rails.groups(:assets)) if defined?(Bundler)
require_relative './configatron.rb'

if configatron.coverband.enabled!
  require 'coverband'
end
```

danmayer: OK, yeah I don't think we want

```
This is in routes.rb

require 'coverband/reporters/web'
```

I don't think we want folks to require in routes anymore... I think we could make it work if they do by duplicating all the extra requires in the normal initialize

danmayer: normal as in the one in web not the new override.

kbaum: Yeah i can totally fix this for this specific installation by not having the includes in the routes but think it would save everyone headaches if we try to make it as backwards compatible as possible.

kbaum:
@danmayer - take a look at this change. If someone requires `coverband/reporters/web` first, we explicitly require 'coverband' first. Also, we don't bother defining `initialize` within the web.rb file since we also define it in coverband.rb. I think this takes away confusion around load order issues.

lib/coverband.rb
```diff
@@ -18,6 +18,7 @@
 require "coverband/collectors/view_tracker"
 require "coverband/reporters/base"
 require "coverband/reporters/console_report"
+require 'coverband/reporters/web'
```
danmayer, line 21: so if we require this it is always loaded, aren't we trying to not have this loaded unless the user tries to instantiate it... if we add this here we end up double requiring this file later in the over-riden init method.

kbaum: Yes that should not be there. Great catch.

kbaum: Fixed!

Issue #399, opened 2020-09-17
View tracking is missing some views


jagthedrummer:
We have a handful of view files that aren't showing up in either the used or unused lists. One such file lives at `app/views/backpack/affiliate_commissions/new.html.haml`. My first suspicion was that it's being missed because it doesn't have an `erb` extension anywhere, but there are other similarly named files in the same directory that do appear in the lists. For instance `app/views/backpack/affiliate_commissions/edit.html.haml ` does show up in the "used views" list. (For the record we susptec that `new.html.haml` is not used and should be in the "unused views" list.)

danmayer:
yeah, I have fixed this sort of thing by creating one off dynamic view trackers... but the default one is still pretty hard coded for ERB

https://github.com/danmayer/coverband/blob/master/lib/coverband/collectors/view_tracker.rb#L17

I would be interested in making this more flexible out of the box. I think I would just need to add a list of other common template lib extensions `haml`, `slim`, `etc`

markshawtoronto:
Same problem for me, in my case also `*.html.haml` 👍

danmayer:
ok folks, check out this PR, it should resolve the issue and explain how one could fix in apps if they want to do custom configuration

https://github.com/danmayer/coverband/pull/401

I will work to get a 5.0.2 release with this in it in a couple of days once I get some feedback

danmayer:
release in a day or two sorry for the delay

danmayer:
fixed in release 5.0.2

PR #401, opened 2020-09-20
Support haml slim by default address issue #399

danmayer:
Ok, yeah basically I have one off set this up in a couple apps but not supported it out of the box until now...

If you setup your own view tracker you can pass in your own 'target' set of files: https://github.com/danmayer/coverband/blob/master/lib/coverband/collectors/view_tracker.rb#L28

Which allows you to go beyond the default ERB, that being said HAML and Slim are really common, so it would be good to support them out of the box.

This builds on another PR which made it easier to test the view tracker... most of this code is testing the real diff is basically

```
- Dir.glob("app/views/**/*.html.erb").reject { |file| file.match(/(_mailer)/) }
+ Dir.glob("#{@project_directory}/app/views/**/*.html.{erb,haml,slim}").reject { |file| file.match(/(_mailer)/) }
```

take a look @jagthedrummer @markshawtoronto @ChristopherHelgeson

### DIFF
--- a/lib/coverband/collectors/view_tracker.rb
+++ b/lib/coverband/collectors/view_tracker.rb
@@ -10,11 +10,8 @@ module Coverband
     # This is a port of Flatfoot, a project I open sourced years ago,
     # but am now rolling into Coverband
     # https://github.com/livingsocial/flatfoot
-    #
-    # TODO: test and ensure slim, haml, and other support
     ###
     class ViewTracker
-      DEFAULT_TARGET = Dir.glob("app/views/**/*.html.erb").reject { |file| file.match(/(_mailer)/) }
       attr_accessor :target, :logged_views, :views_to_record
       attr_reader :logger, :roots, :store, :ignore_patterns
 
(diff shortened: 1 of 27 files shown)


lib/coverband/collectors/view_tracker.rb
```diff
@@ -10,22 +10,20 @@ module Collectors
...
       def initialize(options = {})
         raise NotImplementedError, "View Tracker requires Rails 4 or greater" unless self.class.supported_version?
+        raise "Coverband: view tracker initialized before configuration!" if !Coverband.configured? && ENV["COVERBAND_TEST"] == "test"
 
         @project_directory = File.expand_path(Coverband.configuration.root)
         @ignore_patterns = Coverband.configuration.ignore
         @store = options.fetch(:store) { Coverband.configuration.store }
         @logger = options.fetch(:logger) { Coverband.configuration.logger }
-        @target = options.fetch(:target) { DEFAULT_TARGET }
+        @target = options.fetch(:target) { Dir.glob("#{@project_directory}/app/views/**/*.html.{erb,haml,slim}").reject { |file| file.match(/(_mailer)/) }.freeze }
```
kbaum, line 26: I know this is not related to this PR but does the reject of `_mailer` mean we don't support email templates?

test/forked/rails_full_stack_views_test.rb
```diff
@@ -0,0 +1,48 @@
...
+    Coverband.report_coverage
+  end
+
+  def teardown
+    super
+    Capybara.reset_sessions!
+    Capybara.use_default_driver
+  end
+
+  test "verify erb haml slim support" do
```
kbaum, line 21: Nice tests!

kbaum (approved):

markshawtoronto (approved):
Amazing! Thanks for the tip and the fix, we love coverband 😉

lib/coverband/collectors/view_tracker.rb
danmayer: yeah I can look at it in another PR, I believe in Rails back in the day (Rails 3) we didn't receive active support notifications for email templates... I believe that has been fixed so this could be updated.

PR #406, opened 2020-10-07
Fix for "Unused Views" tracking.

markshawtoronto:
I noticed that the "Unused Views" aren't removing entries that are used when I use it, because they don't go through the same root parsing.

Steps to reproduce:
I made `rails_full_stack_views_test.rb` fail by checking for `Unused Views` count going down when `Used Views` goes up.

❤️ for coverband!

### DIFF
--- a/lib/coverband/collectors/view_tracker.rb
+++ b/lib/coverband/collectors/view_tracker.rb
@@ -63,18 +63,29 @@ module Coverband
         normalized_views = {}
         views.each_pair do |view, time|
           roots.each do |root|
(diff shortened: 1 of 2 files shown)


markshawtoronto:
@danmayer 👋 I noticed this when trying out the new haml tracking.

danmayer (approved):
Thanks this is great @markshawtoronto

danmayer:
Ok this is in latest release https://github.com/danmayer/coverband/releases/tag/v5.0.3

Issue #255, opened 2019-05-08
UI Support for Showing Dead Methods


blakewest:
Hi! We're using Coverband at my work. We have a large, old rails codebase, and are finding tons of dead methods, which is awesome! Thank you! 

Problem is, we have thousands of files, and percent coverage doesn't tell if any of the methods are completely dead. That is, not a single line of the method has been run in prod. It's a huge hassle to open up all of our 2700 files, and manually search through the methods to find totally dead ones, which are the easiest ones to remove. So a little automated help here would be fantastic!

I think it would be really awesome to add support in the UI for surfacing dead methods. There's various ways to do this. One idea (just spitballin' here) is to have a new tab called "Dead Methods" right next to the existing "All Files" tab, and that tab just lists all the dead methods, and links back to their corresponding file when you click on them. 

I haven't dug too deep, but I feel like determining dead methods should be relatively straightforward. We're just talking about finding where between the start and ending definition of a method, not a single line has been covered. 

Would love to hear your thoughts! I'd be happy to help do some work on this if you could point me in the right direction.

danmayer:
Thanks, @blakewest that is a good idea and a cool feature request. I haven't had to add a real parser into the project yet. I think to do this correctly I would need to pull in a parsing library. I am interested in giving it a shot, but I am a bit worried about the complexity it would add to the project.

I will leave this open and hope to take a look but I don't think I will get to it super soon given closing out some of the smaller issues and other features we have on the pipeline. I might be able to get to it in a couple of months, if you or someone on your team wants to give it a shot I am happy to take pull requests and could setup a quick pairing session to help someone get more familiar with the code so they could dig in more quickly.

blakewest:
@danmayer yeah, we may be interested in doing some work on it. A pairing sesh could be good. I took a brief look through some of the view code, but your help would be appreciated for getting a strategy. 

As far as the parsing library goes, I think we can do a V1 without importing a whole library. Since we just want to find dead methods, we can keep it simple. Option 1: Read in files, and just use indentation level. As in, find a line with `def`, then note it's indentation level. Keep reading lines until you see `end` at the same level. A bit hacky, but prob would work like 99% of the time for any reasonable codebase. Alternatively, we can steal the approach of the `method_source` gem. See their `complete_expression?` method [here](https://github.com/banister/method_source/blob/master/lib/method_source/code_helpers.rb#L66). Spoiler, they use eval inside a `catch`.

[...]

PR #410, opened 2020-12-20
Add dead_methods rake task

kbaum:
Add rake task which outputs all dead methods based on current coverage data:

```
+---------------------------------------+-----------------+-------------------------+-------------+
➜  coverband_demo git:(master) ✗ ./bin/rake coverband:dead_methods
+---------------------------------------+-----------------+-------------------------+-------------+
| file                                  | class           | method                  | line_number |
| ./app/controllers/posts_controller.rb | PostsController | edit                    | 22          |
| ./app/controllers/posts_controller.rb | PostsController | update                  | 43          |
| ./app/controllers/posts_controller.rb | PostsController | destroy                 | 57          |
| ./app/controllers/posts_controller.rb | PostsController | destroy_bad_posts       | 65          |
| ./app/controllers/posts_controller.rb | PostsController | destroy_bad_dangerously | 73          |
| ./app/controllers/home_controller.rb  | HomeController  | trigger_jobs            | 8           |
| ./app/controllers/home_controller.rb  | HomeController  | data_tracer             | 14          |
+---------------------------------------+-----------------+-------------------------+-------------+
```

Uses `RubyVM::AbstractSyntaxTree` to find method and class definitions. Works with ruby 2.6+.


danmayer:
Yeah my biggest concern on this is just long term maintenance of this and that the real value as mentioned in the issue is when it is hooked up with CI or the build process like github status integration. Then it can attempt removals and build up cleanup PR... Which is a bit out of scope for the gem. I guess if we can find a way to not need the new dependency I would be OK adding this as something that makes it easier for folks to build off

kbaum:
> the real value as mentioned in the issue is when it is hooked up with CI or the build process like github status integration. Then it can attempt removals and build up cleanup PR

Those features sound very interesting for a service and agree they are out of scope for the gem. That being said, a simple rake task that spits out dead methods might be an easy way to find dead code. We could possibly highlight dead methods in the report as well.

> I guess if we can find a way to not need the new dependency

We don't need that dependency. Just used for formatting the output of the rake task but can do that without a gem.

kbaum:
Removed dependency on terminal-table

lib/coverband.rb
```diff
@@ -6,6 +6,7 @@
 require "coverband/version"
 require "coverband/at_exit"
 require "coverband/configuration"
+require "coverband/utils/dead_methods"
```
danmayer, line 9: do we want these just conditional required in the rake task, etc... since it only supports later Ruby methods wouldn't trying to call refine cause issues?

lib/coverband/utils/tasks.rb
```diff
@@ -9,11 +9,23 @@
     Coverband::Reporters::ConsoleReport.report(Coverband.configuration.store)
   end
 
+  if defined?(RubyVM::AbstractSyntaxTree)
+    desc "Output all dead methods"
+    task :dead_methods do
+      Coverband::Utils::DeadMethods.output_all
```
danmayer, line 15: I might just add the requires for the two things here since this checks for the required support and a user is trying to run it

kbaum: I like that idea. Will do.

kbaum: This is done.

kbaum:
@danmayer Would it make sense to add dead methods to the web UI as mentioned [here](https://github.com/danmayer/coverband/issues/255#issue-441836286)?

danmayer:
> @danmayer Would it make sense to add dead methods to the web UI as mentioned [here](https://github.com/danmayer/coverband/issues/255#issue-441836286)?

Not sure, I am curious how many folks would use this feature if you want to add a UI, I am cool with that, if you want to do a release with the Rake task and see if folks give it a go that is cool too... I would add and update a section on the readme about how to check out this new feature.

danmayer (approved):

kbaum:
>  I would add and update a section on the readme about how to check out this new feature.

Good call. Will do.

kbaum:
Added Newer Features section with dead methods rake tasks.

kbaum:
Think it's ok to do a version bump for this?

danmayer:
> Think it's ok to do a version bump for this?

Since it is non-breaking and additive only, and likely not used by all users, I would do a minor bump... So `5.1.3`...

If you are going to do a web-UI, I might wait and release both the rake and web-ui together.

kbaum:
Ok i'll do a minor version bump. Not sure when I'll get around to do the web UI.

kbaum:
> If you are going to do a web-UI, I might wait and release both the rake and web-ui together.

I would go for the web UI but I think it might be more involved due to performance considerations. Each time we calculate dead methods we need to read every source file in the project. I'm not sure how reasonable that would be on a large project. We could opt for a separate rake task which calculates the dead method data and stores in redis. Then the report UI could just pull from redis.

danmayer:
No worries we can do a gem release with this and if there is a lot of interest we can revisit trying a web UI, or we can see where that might fit into a more robust service ;)

PR #411, opened 2020-12-25
Add support for ruby-3.0

kbaum:
Had to pin to master of redis-namespace to get this change:

https://github.com/resque/redis-namespace/commit/c31e63dc3cd5e59ef5ea394d4d46ac60d1e6f82e

Probably should wait until we get a version bump on that dependency before merging this in.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
         # remove until I sort out CI issues for truffle
         # truffleruby,
         # truffleruby-head,
-        ruby: [2.3, 2.4, 2.5, 2.6, 2.7, jruby, jruby-head]
+        ruby: [2.4, 2.5, 2.6, 2.7, 3.0, jruby, jruby-head]
         redis-version: [5]
     runs-on: ${{ matrix.os }}-latest
     steps:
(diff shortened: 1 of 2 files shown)


danmayer:
Nice, but yeah let's wait for a redis release... Speaking of which we could make reid's optional with the other drivers

kbaum:
@danmayer Got a version bump for redis-namespace. Think this is worth another look.

danmayer:
yeah for sure, if it all passes CI we can at least get this in, I would like to get it running on a full app to be sure things function, but  I don't have any apps on Ruby 3.0 yet.

kbaum:
We could install it on coverband_demo with ruby 3.0?

.github/workflows/main.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         # remove until I sort out CI issues for truffle
         # truffleruby,
         # truffleruby-head,
-        ruby: [2.3, 2.4, 2.5, 2.6, 2.7, jruby, jruby-head]
+        ruby: [2.4, 2.5, 2.6, 2.7, 3.0, jruby, jruby-head]
```
hosamaly, line 20: Is there a reason to drop support for Ruby 2.3? Are there any failing tests?

kbaum: I am not sure why I did that honestly. Opened #417 to add it back.

danmayer:
yeah I think we normally have said we would maintain what is still
supported on heroku... but we also wouldn't drop a version until we had
reason like some feature or failing test... So I am fine with keeping it we
definitely still have users on very old versions of Ruby

Dan Mayer
http://mayerdan.com
http://twitter.com/danmayer

PR #412, opened 2020-12-29
checked this in a branch I have for coverband service and verified th…

danmayer:
…at mailers seem to be picked up as expected now

As discussed a long time ago in this PR (https://github.com/danmayer/coverband/pull/401), ensure we can show email templates, views, and partials for action mailer related view files...


danmayer:
I had to have at_exit also report views to ensure that we now would report emails that were generated by rake tasks or other system jobs like crons... 

I have a test mailer in coverband service and it got picked up along with the mailer layout as a used view.
https://coverband.io/gh/coverband-service/coverband_service/view_trackers?cleanup_mode=1&runtime_env=production

kbaum (approved):
Makes senes. Looks great!

danmayer:
OK, cool I removed the version changes and this can go out when you release the method detection stuff @kbaum

PR #415, opened 2021-01-31
Require Ruby 2.3 or later

hosamaly:
The gem is incompatible with Ruby 2.2 because it uses the safe-navigation operator that was introduced in Ruby 2.3. Support for Ruby 2.2 and older was dropped in v3.0.

### DIFF
--- a/coverband.gemspec
+++ b/coverband.gemspec
@@ -20,6 +20,8 @@ Gem::Specification.new do |spec|
   spec.executables = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.test_files = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = %w[lib]
+  
+  spec.required_ruby_version = '>= 2.3'
 
   spec.add_development_dependency "benchmark-ips"
   spec.add_development_dependency "capybara"


kbaum (approved):
This makes a lot of sense since we are only running tests agains ruby 2.3+.

https://github.com/danmayer/coverband/blob/66f8f1268e864bc41f2fdcf0a4d93dc97ee54be8/.github/workflows/main.yml#L20

Thanks @hosamaly !

danmayer:
cool thanks @hosamaly where you trying to use this on an older Ruby or just noticed? Anyways will merge it in as we definitely aren't trying to support Ruby that past end of life support.

hosamaly:
Thank you, @danmayer and @kbaum.

Indeed, I noticed it in a project that happens to run Ruby 2.2. Thanks for merging it.

danmayer:
cool thanks @hosamaly if you go back to a last major release of coverband at say `2` or perhaps `3`, my guess is it would support 2.2, I forget when we officially dropped support, but if you need it for a really old app that would be my recommendation.

based on the change log we dropped to support at 3.0.0, so a prerelease of 3 or the very last stable release of 2.x would likely still work if you are really in need.

https://github.com/danmayer/coverband/blob/master/changes.md#coverband-300

PR #416, opened 2021-02-12
Test against Redis 4 and Redis 6 in addition to 5

hosamaly:
The `redis` gem should allow us to support all of them.


danmayer:
thanks @hosamaly

PR #417, opened 2021-02-13
Add ruby 2.3 back

kbaum:
I'm not sure why I removed ruby 2.3 from the build. Seems to work fine.


danmayer:
works for me, thanks

PR #418, opened 2021-02-16
Improve ViewTracker performance

hosamaly:
Improve performance of `ViewTracker` by using sets to back its implementation.

Although they may be slightly slower to add keys, they are quicker to check the existence of keys (in `newly_seen_file?`), which should make it faster overall (since keys are added once and checked every time a view is rendered).

This change only affects the inner implementation of the `ViewTracker`. The exposed interface still returns an `Array`.

### DIFF
--- a/lib/coverband/collectors/view_tracker_service.rb
+++ b/lib/coverband/collectors/view_tracker_service.rb
@@ -8,8 +8,8 @@ module Coverband
     class ViewTrackerService < ViewTracker
       def report_views_tracked
         reported_time = Time.now.to_i
-        if views_to_record.any?
-          relative_views = views_to_record.map! do |view|
+        if @views_to_record.any?
+          relative_views = @views_to_record.map! do |view|
             roots.each do |root|
               view = view.gsub(/#{root}/, "")
             end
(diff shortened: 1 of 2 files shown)


lib/coverband/collectors/view_tracker_service.rb
```diff
@@ -8,16 +8,16 @@ module Collectors
     class ViewTrackerService < ViewTracker
       def report_views_tracked
         reported_time = Time.now.to_i
-        if views_to_record.any?
-          relative_views = views_to_record.map! do |view|
+        if @views_to_record.any?
+          relative_views = @views_to_record.map! do |view|
```
hosamaly, line 12: I don't like using instance variables everywhere, but I felt like I had to do this to keep the attribute readers returning an array.

If it's ok to change their type in the exposed API, then I think that it would be better to undo the use of instance variables directly (as long as the whole code can still work with sets).

danmayer: yeah, gotcha... I guess since other interfaces call and use it, it is the safest change to do it this way avoiding breaking any external users... I am good with it.

danmayer (approved):
looks good to me, I will get this in the next release... We are just doing some tests against Ruby 3.0 on a deployed app to ensure we didn't miss anything.

PR #425, opened 2021-08-18
Improve View tracker to track render collection too

stoivo:
Coverband should track render template, partial and collection, each of these would push message `render_template.action_view`, `render_partial.action_view`, `render_collection.action_view`

### DIFF
--- a/lib/coverband/utils/railtie.rb
+++ b/lib/coverband/utils/railtie.rb
@@ -37,7 +37,7 @@ module Coverband
 
           Coverband.configuration.view_tracker = COVERBAND_VIEW_TRACKER
 
-          ActiveSupport::Notifications.subscribe(/render_partial.action_view|render_template.action_view/) do |name, start, finish, id, payload|
+          ActiveSupport::Notifications.subscribe(/render_(template|partial|collection).action_view/) do |name, start, finish, id, payload|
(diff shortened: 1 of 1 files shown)


danmayer (approved):

danmayer:
@stoivo this is in gem 'coverband', '= 5.2.0.rc.1' if you want to help test the RC

stoivo:
There is anouther issue, i dont have the view tracker tab at all. I will try to figure out why

stoivo:
bisect it down to 11ec00d344d64ec624e21a9df9bd71efd4f1017e

stoivo:
Turns out I didn't set COVERBAND_ENABLE_DEV_MODE to true, should I need to set it? Add a comment to readme

danmayer:
hmmm no you shouldn't let me take a closer look at that commit thanks @stoivo

danmayer:
cool so this bug only impacted development, not production, but still awesome to find and fix it in the RC release... There is a new release 'coverband', '= 5.2.0.rc.2' if you want to give that a shot @stoivo and thanks so much

stoivo:
Now, its working in development and staging.

stoivo:
We have deploy it and it works perfectly. Thankgs for merging and maintaining coverband ❤️

PR #427, opened 2021-09-09
Fix disabling views tracking

dostu:
If I set `config.track_views = false` in configuration, `track_views` method still returns `true`. 

The issue was introduced with this commit: https://github.com/danmayer/coverband/commit/0d82c5cc12449fbfa9f5577b7cea31243a9e3460

### DIFF
--- a/lib/coverband/configuration.rb
+++ b/lib/coverband/configuration.rb
@@ -147,7 +147,9 @@ module Coverband
     end
 
     def track_views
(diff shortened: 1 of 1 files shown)


danmayer:
awesome thanks for the fix!

danmayer:
@dostu this is in gem 'coverband', '= 5.2.0.rc.1' if you want to help test the RC

PR #433, opened 2021-12-16
Refactor filling zero coverage for not loaded files

kei-s:
`Coverband::Utils::Result.add_not_loaded_files` method counts number of lines of not loaded files and fill zero as coverage.
However, `File.foreach(absolute).count` would be same result of `Coverband::Utils::SourceFile#src`'s number of lines.
`File.foreach` or `File.read` are heavy processes. This results too slow response when app has some not loaded files.

This PR fix that duplicated processes.

### DIFF
--- a/lib/coverband/utils/result.rb
+++ b/lib/coverband/utils/result.rb
@@ -55,7 +55,7 @@ module Coverband
           Dir[tracked_files].each do |file|
             absolute = File.expand_path(file)
             result[absolute] ||= {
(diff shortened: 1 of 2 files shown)


danmayer:
thanks, I will review a bit more closely soon

danmayer:
@kei-s this is in gem 'coverband', '= 5.2.0.rc.1' if you want to help test the RC

PR #438, opened 2022-01-25
Fix deprecated uses of `Redis#pipelined`

casperisfine:
Context: https://github.com/redis/redis-rb/pull/1059

The following is deprecated
```ruby
redis.pipelined do
  redis.get(key)
end
```

And should be rewritten as:
```ruby
redis.pipelined do |pipeline|
  pipeline.get(key)
end
```

Functionally it makes no difference.
This API is available since Redis 3.0.

### DIFF
--- a/lib/coverband/adapters/hash_redis_store.rb
+++ b/lib/coverband/adapters/hash_redis_store.rb
@@ -88,9 +88,9 @@ module Coverband
 
       def coverage(local_type = nil)
         files_set = files_set(local_type)
-        @redis.pipelined {
-          files_set.map do |key|
-            @redis.hgetall(key)
+        @redis.pipelined { |pipeline|
+          files_set.each do |key|
+            pipeline.hgetall(key)
           end
         }.each_with_object({}) do |data_from_redis, hash|
           add_coverage_for_file(data_from_redis, hash)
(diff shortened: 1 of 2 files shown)


lib/coverband/adapters/hash_redis_store.rb
```diff
@@ -88,9 +88,9 @@ def save_report(report)
 
       def coverage(local_type = nil)
         files_set = files_set(local_type)
-        @redis.pipelined {
-          files_set.map do |key|
-            @redis.hgetall(key)
+        @redis.pipelined { |pipeline|
+          files_set.each do |key|
```
danmayer, line 92: any reason the `map` moved to an `each`?

byroot: Yes because the return value doesn't matter. `pipelined` doesn't care about what the block returns.

danmayer: gotch, yep all seems good.

danmayer (approved):
merged thanks

danmayer:
oops sorry merged... I will try to get a release out in the next day or two @natematykiewicz / @casperisfine

casperisfine:
No big deal It's just a warning and people have a flag to silence them.

danmayer:
OK, sorry for the delay, this has been released in 5.2.2

PR #440, opened 2022-02-28
 Fixed "TypeError" in ignore

penguinwokrs:
Hi there

I followed the README to set up ignore and got an error.

https://github.com/danmayer/coverband/blob/main/lib/coverband/configuration.rb#L173

As per README, the variable has not been initialized to an array yet, so I suggest a way to make it work.

error

```
coverage failed to store
Coverband Error: #<TypeError: no implicit conversion of Symbol into Integer> no implicit conversion of Symbol into Integer
```


penguinwokrs:
The cause may have been something else, so I'll close it.

danmayer:
OK, if you have more issues feel free to re-open and add some additional details around the issue

PR #446, opened 2022-04-25
Add a known issue for the Elastic APM ruby agent

lewiseason:
I've just started using Coverband on a project which also uses [Elastic APM](https://github.com/elastic/apm-agent-ruby) and resque, and my jobs were failing. Presumably being caused by a similar problem as the Scout APM issue, and I got a hint from the readme section about Working with environment variables.

Loading coverage before elastic-apm appears to fix the issue.

Thanks for coverband, it's awesome ❤️


danmayer (approved):
thanks yeah, load order is important for coverband, we have some other sections about that in the readme, but happy to call it out specifically here. Thanks

PR #447, opened 2022-04-26
Adds report-only CSP

jwg2s:
We had numerous issues getting the Coverband web UI to work with our existing CSP rules. This isn't perfect but is an example of how you could fold it into the existing web.rb file.

Feel free to close this and replace with something where you could change it from `Content-Security-Policy-Report-Only` to `Content-Security-Policy` instead.

### DIFF
--- a/lib/coverband/reporters/web.rb
+++ b/lib/coverband/reporters/web.rb
@@ -13,11 +13,27 @@ module Coverband
     class Web
       attr_reader :request
 
+      CSP_HEADER = [
+        "default-src 'self' https: http:",
+        "child-src 'self'",
+        "connect-src 'self' https: http: wss: ws:",
+        "font-src 'self' https: http:",
+        "frame-src 'self'",
+        "img-src 'self' https: http: data:",
+        "manifest-src 'self'",
+        "media-src 'self'",
+        "object-src 'none'",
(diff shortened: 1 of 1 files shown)


jwg2s:
If you give some guidance, I'm happy to fix the formatting issues and push up an additional commit to reach a CSP you'd be comfortable merging.

danmayer:
No worries on the formatting, I believe we have a checked in standardrb config, so if you run `standardrb --fix` it should match the existing style... If that doesn't work for you I can fix up formatting after the PR is good to go and merged.

I think since this hasn't been an issue for most folks and many projects it would be good to make it optional could we have a configurable option off by default. `Coverband.configuration.csp` then if true we add in the headers? Would that work for your needs? @jwg2s

danmayer:
hey folks sorry for the delay on this got a bit lost in my queue and we had an alternative solution in our apps so I hadn't needed anything like this... Anyways, I saw other were pulling this in forks so I cleaned it up on another branch and got a configuration option and it is merged into main... you can give it a go with

`5.2.6.rc.3` to enable CSP add this to your coverband config file

`config.csp_policy = true`

danmayer:
let me know if this works @jwg2s or @scruti

danmayer:
full details on the code change https://github.com/danmayer/coverband/pull/470

PR #458, opened 2022-08-17
Fix specs for Ruby 3.1 by skipping Simplecov

anonoz:
Ruby 3.1 has made some [breaking changes](https://bugs.ruby-lang.org/issues/18176) to Coverage API, which causes Simplecov to not work well with Coverband together.

This PR:

- skips Simplecov for any Ruby >= 3.1 as a temporary fix.
- patches `HashRedisStore` for using ActiveSupport's `blank?` method, which somehow does not get loaded in Ruby 3.1. This is not a proper solution but I do think it would be nice to not use any ActiveSupport/Rails APIs unless necessary.
- Add ruby 3.1 to CI test matrix.

### DIFF
--- a/lib/coverband/adapters/hash_redis_store.rb
+++ b/lib/coverband/adapters/hash_redis_store.rb
@@ -119,7 +119,7 @@ module Coverband
 
         data = coverage_data_from_redis(data_from_redis)
         hash[file] = data_from_redis.select { |meta_data_key, _value| META_DATA_KEYS.include?(meta_data_key) }.merge!("data" => data)
-        hash[file][LAST_UPDATED_KEY] = hash[file][LAST_UPDATED_KEY].blank? ? nil : hash[file][LAST_UPDATED_KEY].to_i
+        hash[file][LAST_UPDATED_KEY] = (hash[file][LAST_UPDATED_KEY].nil? || hash[file][LAST_UPDATED_KEY] == '') ? nil : hash[file][LAST_UPDATED_KEY].to_i
         hash[file].merge!(LAST_UPDATED_KEY => hash[file][LAST_UPDATED_KEY], FIRST_UPDATED_KEY => hash[file][FIRST_UPDATED_KEY].to_i)
       end
 
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -1,11 +1,22 @@
 # frozen_string_literal: true
 
+require 'bigdecimal'
+
 original_verbosity = $VERBOSE
 $VERBOSE = nil
+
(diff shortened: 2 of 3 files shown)


lib/coverband/adapters/hash_redis_store.rb
```diff
@@ -119,7 +119,7 @@ def add_coverage_for_file(data_from_redis, hash)
 
         data = coverage_data_from_redis(data_from_redis)
         hash[file] = data_from_redis.select { |meta_data_key, _value| META_DATA_KEYS.include?(meta_data_key) }.merge!("data" => data)
-        hash[file][LAST_UPDATED_KEY] = hash[file][LAST_UPDATED_KEY].blank? ? nil : hash[file][LAST_UPDATED_KEY].to_i
+        hash[file][LAST_UPDATED_KEY] = (hash[file][LAST_UPDATED_KEY].nil? || hash[file][LAST_UPDATED_KEY] == '') ? nil : hash[file][LAST_UPDATED_KEY].to_i
```
danmayer, line 122: Yeah, a good call we do try to avoid Rails methods, this slipped in at some point... I guess most of our users are on Rails so we missed it.

test/test_helper.rb
```diff
@@ -1,11 +1,22 @@
...
+
+if ENV['SKIP_SIMPLECOV'] || BigDecimal(RUBY_VERSION[0, 3]) >= BigDecimal('3.1')
+  $SKIP_SIMPLECOV = true
+end
+
 require "rubygems"
 require "pry-byebug" unless ENV["CI"]
-require "simplecov"
-require "coveralls"
+
```
danmayer, line 14: Do you have any more info on the issue you were seeing related to simplecov? I don't see immediately why the changes to Coverage would cause an issue.

anonoz: The changes makes `Coverage#start` raise an error on 2nd invocation.

See https://github.com/ruby/ruby/pull/4856/files#diff-1dc6d1837e2b8b7d8b1ae5a3067b3a677904c72dd31d0ca0b95d20f6691f7510R47

In ruby 3.0 and older, we can keep calling `Coverage#start` without raising exception.

danmayer: ahh perfect thanks, I can get a test around this and check if enabled and not call or capture the exception. Thanks... Feel free to test off the mainline branch or I can cut an RC sometime in next day or two after fixing this exception.

danmayer: ok fixed here: https://github.com/danmayer/coverband/commit/2ce9355ccc3c8153a72214b19f8c6d7c857881cb

danmayer:
this is live in a RC release `5.2.5.rc.1` if you want to try that in your gemfile, I will make a full release in a day or two after testing in some apps

PR #459, opened 2022-10-01
Route tracker

danmayer:
This adds support for route tracking based on an active support event. This event is on main and will be in the next release of Rails. I am going to look to back port support for this event for prior versions of rails. 

---

This closely follows the pattern established for the view tracker, and with a follow up that will add translation key tracking will hit the rule of threes and I will abstract out how to add tracking for any active support notification events.

---
screenshots (blocked out private repo info):

Screen Shot 2022-10-01 at 4 15 04 PM
Screen Shot 2022-10-01 at 4 15 10 PM


lib/coverband/reporters/web.rb
```diff
@@ -126,7 +140,7 @@ def clear
         else
           notice = "web_enable_clear isn't enabled in your configuration"
         end
-        [301, {"Location" => "#{base_path}?notice=#{notice}"}, []]
```
danmayer, line 129: note these should have always been 302s, just realized it while building this out

colemanja91 (approved):
There's a fair amount of Coverband-specific code here I'm not super familiar with, but the following looks good:

* Defining what is a "unique route"
* Integration with ActiveSupport notifications
* Reporting on used vs. unused

PR #472, opened 2023-03-10
Fix service_test_mode configuration

rileyanderson:
~`service_disabled_dev_test_env?` should return true if there is no service.~

~We've noticed an influx of Coverband errors in our test suite since this was added. - `ERROR -- : Coverband: view_tracker~ ~failed to store, error NoMethodError`~

~I believe this check should return true. If there is no service, it can be considered disabled. Is my thinking correct on this?~

Fixes `@service_test_mode` configuration

### DIFF
--- a/lib/coverband/configuration.rb
+++ b/lib/coverband/configuration.rb
@@ -272,7 +272,7 @@ module Coverband
     end
 
     def service_test_mode
-      @service_dev_mode ||= ENV["COVERBAND_ENABLE_TEST_MODE"] || false
+      @service_test_mode ||= ENV["COVERBAND_ENABLE_TEST_MODE"] || false
     end
 
     def process_type


danmayer:
hmm this is more likely that it doesn't have access to a redis in the test environment or test setup. If you have coverband in the test environment and are running it would expect to be able to store things...

If you are talking about something different in terms of "is no service" you mean running coverband outside of a webapp or something without views, could you give me more details?

danmayer:
Also, in case it helps we generally don't recommend coverband for testing and point folks to `simplecov` it is more intended for production, staging, and other runtime environments than for running tests, although some folks find it helpful for running selenium style tests to exercise a server.

rileyanderson:
Ahh ok that makes sense. I guess I had wrongly assumed that it was supposed to be disabled by default in test.

Would you still like [this change](https://github.com/danmayer/coverband/pull/472/files?diff=unified&w=0#diff-7156f65e8150c9767d87e2f12c84303875e74347116011e59b5e0700deb0f469R274-R276)? I can throw it up another PR if you'd prefer.

danmayer:
yeah I will take the test mode change if you want to just remove the other change from this PR, thanks

rileyanderson:
> yeah I will take the test mode change if you want to just remove the other change from this PR, thanks

Updated. Thanks for your help!

PR #475, opened 2023-03-28
Fix minitest deprecation warning (`assert_nil`)

etagwerker:
Hi there,

It seems that Minitest 6 will no longer support this idiom: `assert_equal nil, obj`:

> DEPRECATED: Use assert_nil if expecting nil from /Users/etagwerker/Projects/fastruby/coverband/test/coverband/collectors/route_tracker_test.rb:23. This will fail in Minitest 6.

So this PR fixes that problem and makes it easier to upgrade to Minitest 6.

Thank you!


danmayer (approved):
thanks so much

PR #476, opened 2023-04-19
Enable or disable Coverband through option in Configuration

Rukomoynikov:
Hello folks, and thank you for such a great piece of software.

In our project, we don't use ENV variables for configuration. But we want to enable Coverband only for the production environment and disable it for development and test environments. 

As I understand, in the Rails world, it's pretty common to have configuration files where we can use something like this:

```
GemName.configure do |config|
  config.enable = Rails.env.production? || Rails.env.staging?
end
```

### DIFF
--- a/lib/coverband.rb
+++ b/lib/coverband.rb
@@ -103,7 +103,7 @@ module Coverband
     Coverband::Collectors::Coverage.instance
   end
 
-  unless ENV["COVERBAND_DISABLE_AUTO_START"]
+  if Coverband.configuration.enabled
     begin
       # Coverband should be setup as early as possible
       # to capture usage of things loaded by initializers or other Rails engines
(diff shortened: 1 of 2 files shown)


danmayer:
I guess we normally just recommend using `require: false` in the Gemfile if you don't want to use ENV vars then in something like an initializer or production.rb you can `require 'coverband'`

I guess if we you think the config option via the configuration file is easier it could be done but it will load and run some stuff only to short circuit later... might as well just not load coverband unless you want to.

danmayer:
to avoid any configuration bugs I do generally recommend running coverband in development, but I often do not include it for test

Rukomoynikov:
Thank you @danmayer 

I agree `require false` in Gemfile and then manual requiring in `config/environments/production.rb` is working well. I will close the PR.

PR #495, opened 2023-12-06
[Proposal] Introduce send_deferred_eager_loading_data configuration to reduce the load on Redis during deployment

makicamel:
This is another way to avoid cache stampede by omitting some reporting on starting servers.
Coverband stores the results of eager_loading to Redis at server startup.
The eager_loading results are big and the same for all servers, so there is no need to save all results.
(In our project we have hundreds of servers, so even with `reporting_wiggle` set to 400, Redis hung.)
By configuring the eager_loading results of some servers to be stored in Redis, we can reduce the load on Redis during deployment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,11 +252,24 @@ end
 
 ### Avoiding Cache Stampede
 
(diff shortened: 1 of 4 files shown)


danmayer (commented):
thanks yeah this makes sense and I could see it causing issues in some cases. Thanks for the PR

danmayer:
OK, I will take a closer look at this and others to run soon

makicamel:
@danmayer 
Thank you for your comment.
I know you are busy, but I would appreciate it if you could take a look :)

danmayer:
hey sorry I lost track of this let me try to get to it tomorrow.

danmayer:
hey sorry again about the delay. I really appreciate the nudge and your support.

PR #496, opened 2023-12-17
docs: update README.md to have track_views default as true

khaled-badenjki:
As per the code here
https://github.com/danmayer/coverband/blob/main/lib/coverband/configuration.rb#L72

and after testing it locally, the default value for `track_views` is true, while in the `README.md` file it says it's false.

In this PR i'm trying to resolve this problem


danmayer (approved):

danmayer:
thanks for improving the docs

PR #500, opened 2024-01-23
[Optimization] use memoized hashes in `Results` to get better lookup times

Drowze:
Running Coverband with a significantly big project (I'm currently working on one wtih ~4600 files) may get quite slow. While profiling the project, I found that one big culprit is the `Results` class, where there's a lot of O(n) lookup operations that could but are not currently being memoized.

This PR adds such optimization: replacing `find` operations for hash lookups (so only the first lookup of `eager` and `runtime` type results is slow - due to building the lookup hash)

This optimization alone brought the following load improvement to the coverage report page:
```ruby
# NOTE: running on a HashRedisStore with ~4600 files
# command used:
Benchmark.measure do
  Coverband::Reporters::Web.new.tap do |cov|
    cov.instance_variable_set(:@request,Rack::Request.new(Rack::MockRequest.env_for("/")))
  end.index
  nil
end

# before
# => #<Benchmark::Tms:0x00007f74bf2ee240 @cstime=0.0, @cutime=0.0, @label="", @real=33.93276486300056, @stime=3.5702559999999997, @total=34.058608, @utime=30.488352>

# after
# => #<Benchmark::Tms:0x00007efe452aa248 @cstime=0.0, @cutime=0.0, @label="", @real=25.779949678000776, @stime=3.301538, @total=25.864683, @utime=22.563145>
```

Also tested with both this and #499
```ruby
# with both this and #499
# => #<Benchmark::Tms:0x00007f1374de55b0 @cstime=0.0, @cutime=0.0, @label="", @real=8.71669229300096, @stime=0.7920630000000002, @total=8.723904, @utime=7.931840999999999>
```

### DIFF
--- a/lib/coverband/utils/results.rb
+++ b/lib/coverband/utils/results.rb
@@ -17,7 +17,11 @@ module Coverband
       def file_with_type(source_file, results_type)
         return unless get_results(results_type)
 
-        get_results(results_type).source_files.find { |file| file.filename == source_file.filename }
+        @files_with_type ||= {}
+        @files_with_type[results_type] ||= get_results(results_type).source_files.map do |source_file|
+          [source_file.filename, source_file]
+        end.to_h
+        @files_with_type[results_type][source_file.filename]
       end
 
       def runtime_relevant_coverage(source_file)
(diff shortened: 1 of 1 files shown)


danmayer:
Nice, this is a great improvement and clean

danmayer:

PR #505, opened 2024-02-14
Added dependabot for GitHub Actions and bundler

ydah:
How about using dependabot in this way?
When this works, a pull request is created as follows:
- ruby/bigdecimal#242

It was created with reference to the following:
- ruby/csv#265

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "bundler"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


danmayer (approved):

ydah:
memo) CI failures are fixed with the following commits
https://github.com/danmayer/coverband/pull/504/commits/8bd6e36d4e49ccd87a857b878a443906c14c4f84

danmayer (approved):

PR #506, opened 2024-02-16
ability to hide settings view

jjb:
Thanks for an amazing project!

We would like the ability to hide the settings view.

I haven't tested this - wanted to simultaneously float the idea, and the start of a possible implementation.

Let me know what you think!


danmayer:
yeah we could hide that if you want, just a new optional config

jjb:
okay, updated the PR. we'll try it out in staging soon.

danmayer:
cool I am modifying a big branch for the next release so I might not merge this in until I get that settled, but I can include this in the next release.

danmayer (approved):

PR #513, opened 2024-02-19
Change the year of the license to the first publication date only

ydah:
This PR change the year of the license to the first publication date only.

>DO YOU CHANGE THE YEARS STATED ON YOUR BLOG AND WEBSITE?
>
>Use the earliest date of publication. For example, if you first published on your blog in 2012, when you amend the blog content or add to it, continue to use the 2012 date or use a range of dates such as 2012–2022.

https://www.copyrightlaws.com/copyright-symbol-notice-year/


danmayer:
ah thanks this is nice

Issue #312, opened 2019-08-14
What do the report columns mean?


jjb:
Is it documented somewhere what each report column means? Let me know what they are and I'll make a PR adding the info to the readme. Thanks!

kbaum:
Great point.

- **% covered** - Percentage of total relevant lines covered
- **% runtime** - Percentage of the runtime lines covered where runtime lines are lines that are hit after the application has been eagerly loaded
- **Lines** - total lines in the file including lines unreachable or uncover-able. An unreachable line would be an empty line with no code.
- **Relevant lines** - lines that are coverable - not empty
- **Lines runtime** - Total lines minus uncoverable lines minus the lines that are only hit during eager loading of application
- **Lines missed** - relevant lines not covered
- **Avg hits/line** - Total of coverage to the file divided by relevant lines. 

@danmayer - correct me if I am off on any of these.

We might want to revisit these at some point. Would be interesting to know what people feel is relevant to them.

danmayer:
yeah I guess a few thoughts on how to make this easier to understand.... add the above to the readme? Add hover tool tips?

jjb:
Wow, thanks for all the info!

What does coverable mean? I think I know what reachable means, but just in case, could you define that as well.

kbaum:
I am using the words reachable and coverable synonymously here.

jjb:
haha, gotcha!

So does reachable/coverable means that coverband is able to look at it?

I thought that reachable meant if it was possible for the code to ever be executed? e.g.

```ruby
if thing
  …
elsif !thing
  …
else
  # this code can never be executed
end
```

kbaum:
Coverband depends on ruby's coverage library which let's you know if there's any code on a line. Lines that have no code (empty or just comments) are not "reachable".

The white lines in this picture are not "reachable".

![image](https://user-images.githubusercontent.com/96786/63116863-a1965a00-bf68-11e9-98af-3a5997d25a91.png)

Looks like `end` statements are not reachable either.

jjb:
thanks! I started this wiki doc https://github.com/danmayer/coverband/wiki/Dictionary

jjb:
> The white lines in this picture are not "reachable".

what are the red lines? code that hasn't been executed at runtime?

Are red lines always covered/reachable? If so, it seems like % runtime is pretty much the only important metric? Maybe it should go furthest to the left?

kbaum:
Green is covered lines
Red lines are reachable but not covered.

[...]

PR #522, opened 2024-03-28
add explanation of columns to readme

trivett:
Added some more information to help new users get acquainted with what the web UI means. Mostly ripped right from the words of @kbaum in https://github.com/danmayer/coverband/issues/312


danmayer:
awesome thanks

PR #524, opened 2024-04-07
Fix some typos

ydah:


danmayer (approved):
thanks, these are in a pulled in file so if we pull in again issues might come back if the various JS projects don't fix upstream

PR #528, opened 2024-04-15
Fix route tracker

alpaca-tc:
The routes_tracker had a bug that caused a significant increase in the number of keys, which has been fixed in this PR.

- At least in the Rails 7.0, `payload[:request]` is always true. And if the value of `url_path: payload[:request].path` contained an id, there was a bug that generated a large number of keys for each target id.
  - Perhaps this code expects to branch on redirect. Therefore, it was modified to `payload.key?(:location)`.
- Added option to not track all redirects.
  - For redirects, using conditions such as `get '*pages', redirect: '/'` in `routes.rb` could lead to a huge number of keys. By choosing not to track redirects, a significant increase in the number of keys can be prevented.
- Tiny refactoring. Removed `if route` because of it is always true.

---

Thanks for this great gem! Our application has reduced our code by a few percent 👍

### DIFF
--- a/lib/coverband/collectors/route_tracker.rb
+++ b/lib/coverband/collectors/route_tracker.rb
@@ -27,6 +27,8 @@ module Coverband
       def track_key(payload)
         route = if payload.key?(:location)
           # For redirect.action_dispatch
+          return unless Coverband.configuration.track_redirect_routes
+
           {
             controller: nil,
             action: nil,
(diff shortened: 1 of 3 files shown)


danmayer:
ok thanks @alpaca-tc I will probably need to run this against a couple rails versions and ensure we stay compatible, but I can give that a go.

danmayer:
awesome thanks I merged this in

PR #532, opened 2024-04-25
Override default Configuration root with an env var

captainkevron:
First off, thank you for maintaining this project. It has been extremely valuable for our organization in removing dead code from legacy applications.

We are currently integrating `coverband` with a Sinatra app and ran into a blocker. Due to a pile of shell scripts and symbolic links in our infrastructure, we need to be able to set the configuration `root` via an environment variable. This PR will set the configuration root to the value of the environment variable "COVERBAND_ROOT" and fall back to the previous behavior when it is not set.

For some additional context, our applications are deployed using conventions similar to [Capistrano](https://capistranorb.com/documentation/getting-started/structure/). 

The live code is referenced via a symbolic link `current` pointing to the directory where the code actually lives.

### DIFF
--- a/lib/coverband/configuration.rb
+++ b/lib/coverband/configuration.rb
@@ -60,7 +60,7 @@ module Coverband
     end
 
     def reset
-      @root = Dir.pwd
+      @root = ENV["COVERBAND_ROOT"] || Dir.pwd
       @root_paths = []
       @ignore = IGNORE_DEFAULTS.dup
       @search_paths = TRACKED_DEFAULT_PATHS.dup


lib/coverband/configuration.rb
```diff
@@ -60,7 +60,7 @@ def initialize
     end
 
     def reset
-      @root = Dir.pwd
+      @root = ENV["COVERBAND_ROOT"] || Dir.pwd
```
captainkevron, line 63: Alternatively, if you did not want to introduce an additional env var, we could do:

```
@root = ENV["PWD"] || Dir.pwd
```

The difference is that `ENV["PWD"]` will return the name of the symoblic link, while `Dir.pwd` will return the path of the underlying directory.

danmayer:
I believe this has come up in the past and there are some other ways you can handle this... I am not opposed to pulling this in if you really need it but can you first try using some of the other methods to set the root via existing config options.

inside your config, I think you can try a couple things...

```
# config/coverband.rb NOT in the initializers
  config.root = "/whatever/path"

  # alternatively there was a way to handle multiple / dynamic roots via root_paths array which setups up searching multiple paths
  config.root_paths = ["running_directory/path", "dynamic_path/pattern", "red/blue/paths"]
end
```
```

danmayer:
thanks for the feedback and PR and let me know if any of these options work for you.

captainkevron:
@danmayer Thanks for the suggestion. Setting `config.root` worked fine. That solution was staring me right in the face and I totally missed it. Appreciate the help.

danmayer:
excellent glad that worked for you, I haven't had a need for the changing deploy directory for a long time so I am glad it is still working as expected. Thanks

PR #541, opened 2024-06-14
better hide_settings documentation

jjb:
whoops, in https://github.com/danmayer/coverband/pull/506/ i accidentally pasted the default into the docs instead of showing how to use the setting


danmayer (approved):
ha, I didn't notice that... thanks

PR #552, opened 2024-10-10
Only print "using default configuration" in verbose mode

jamiecobbett:
This gets printed every time I start a rails process, and creates noise without any value.

Every other call to log is covered by this condition (some have others) so let's do that here too.


danmayer (approved):

jamiecobbett:
Thanks @danmayer 👍

danmayer:
thanks this and other things are in the 6.1.3 release https://github.com/danmayer/coverband/releases/tag/v6.1.3

PR #556, opened 2024-10-11
Give background thread a name

fatkodima:
It will be easier to debug in #555, if the thread had a name.


danmayer:
Thanks @fatkodima  merging, I will try to test some of the recent PR merges and get a release out early next week... I don't run sidekiq swarm in any of the various environments I work in, would it be best to cut an RC release and have you confirm it fixes what you need?

PR #559, opened 2024-10-13
Reduce memory allocated when rendering the index page

fatkodima:
When rendering an index page, coverband makes a lot of unneeded allocations.

Before
```
Total allocated: 349.10 MB (1788664 objects)
```

After
```
Total allocated: 212.79 MB (1134381 objects)
```

So, it is now 40% less in my case.

### DIFF
--- a/lib/coverband/configuration.rb
+++ b/lib/coverband/configuration.rb
@@ -61,7 +61,7 @@ module Coverband
     def reset
       @root = Dir.pwd
       @root_paths = []
-      @ignore = IGNORE_DEFAULTS.dup
+      @ignore = IGNORE_DEFAULTS.map { |ignore_str| Regexp.new(ignore_str) }
       @search_paths = TRACKED_DEFAULT_PATHS.dup
       @verbose = false
       @reporter = "scov"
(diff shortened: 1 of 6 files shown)


lib/coverband/configuration.rb
```diff
@@ -222,8 +222,8 @@ def search_paths=(path_array)
     # Don't allow the ignore to override things like gem tracking
     ###
     def ignore=(ignored_array)
-      ignored_array.map { |ignore_str| Regexp.new(ignore_str) }
```
fatkodima, line 225: Here was an attempt to convert keys to regexps, but because of the bug, this line was a no-op.

lib/coverband/configuration.rb
```diff
@@ -222,8 +222,8 @@ def search_paths=(path_array)
...
-      @ignore = (@ignore + ignored_array).uniq
+      ignored_array = ignored_array.map { |ignore_str| Regexp.new(ignore_str) }
+      @ignore |= ignored_array
```
danmayer, line 226: so I think we want to assign here with a uniq... the purpose of this is that folks can add overrides to ignores... traditionally those have always been strings... so it is good to fix the bug, but I think we still need the `(@ignore + ignored_array).uniq` type logic.

danmayer (changes requested):
I think one issues on how ignore is extended

danmayer:
thanks it is great to get performance and memory enhancing PRs, I appreciate all the effort digging in on this @fatkodima

lib/coverband/configuration.rb
fatkodima: `@ignore |= ignored_array` is the same as `@ignore = (@ignore + ignored_array).uniq`. 
Correct me if I am wrong.

danmayer: ha you are correct, I just don't use the single `|=` very often... this works

danmayer:
this and all the other changes you submitted @fatkodima are in the 6.1.3 release https://github.com/danmayer/coverband/releases/tag/v6.1.3

fatkodima:
Thank you! ❤️ Was waiting for it, going to try tomorrow.

PR #561, opened 2024-10-18
Do not swallow `config.ignore` regexp errors

fatkodima:
Currently, when someone passes an invalid regexp string to `config.ignore`, the error is "silently" ignored and reported to the logger. But this is a programmer error and a small chance anyone will look into the log file and notice this error line there.

Better is to loudly raise and fail.


danmayer:
hmmm while I kind of agree with this we might need to hold it until a larger release version as I could see many folks having bad configs that they were never aware of.

danmayer:
this will go in the next major bump

PR #562, opened 2024-10-18
Mention `rake coverband:clear_tracker` in the README

fatkodima:
It was surprising to find, that `rake coverband:clear` does not clear trackers data and there is a separate task for it (which is not mentioned anywhere 😅).

I would prefer for `clear` task to also clear the trackers data, but at least mentioning that task in the README will be good.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,11 @@ Now that Coverband uses MD5 hashes there should be no reason to manually clear c
 
 This can also be done through the web if `config.web_enable_clear` is enabled.
 
+**NOTE**: The previous task does not clear the trackers data (views, routes, translations, etc).
+To clear trackers data, run
+
+`rake coverband:clear_tracker`
+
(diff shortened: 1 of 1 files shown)


danmayer:
good point, could we actually add a clear all that calls all clears, then still have the specific tasks for clearing the coverage and the views, etc...

Then we could also rename the tasks to make more sense in a major release if changing names or adding a namespace, would make things more clear.

danmayer:
also if we want to save all that for later, I am happy to merge this as it is, just let me know your preference @fatkodima

fatkodima:
I think it is ok to merge as it is, and the for a new major version release to have something like:
`coverband:clear_coverage` - clears only coverage
`coverband:clear_tracker` - clears only views etc
`coverage:clear`  - clears all

I can create a separate PR (to not forget) that will be sitting until new major release.

danmayer:
docs in 6.1.4

PR #564, opened 2024-10-22
Load trackers for `clear_tracker` rake task

fatkodima:
Currently, `clear_tracker` task is a no-op, because `Coverband.configuration.trackers` returns an empty array.
We need to explicitly load rails-related trackers.


danmayer (approved):
thanks

PR #565, opened 2024-10-23
Consider eager loaded code too when detecting dead methods

fatkodima:
Currently, if the file was loaded during eager loading and then its code is never executed during runtime, then it will not have any runtime coverage and when printing dead methods, this file will be missed. 

When reporting dead methods, we need to look at all the files discovered during the eager loading and runtime phases.

### DIFF
--- a/lib/coverband/utils/dead_methods.rb
+++ b/lib/coverband/utils/dead_methods.rb
@@ -37,7 +37,11 @@ module Coverband
       end
 
       def self.scan_all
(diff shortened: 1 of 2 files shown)


danmayer (approved):
good find, I haven't worked with the dead code auto detection part of the project for awhile. If you are using it and find it useful let me know. I  used to have some scripts to turn detected dead code into PRs, but it sometimes found false positives... generally admin or seasonal code that was just rarely used.

PR #566, opened 2024-11-15
Drop support for older rubies, rails and redis

fatkodima:
Also a candidate for a new major release, imo.

This PR drops support for: 
1. ruby < 3.1 (https://devcenter.heroku.com/articles/ruby-support-reference#supported-runtimes, https://endoflife.date/ruby) 
2. rails < 7 
3. redis < 6 (https://redis.io/docs/latest/operate/rs/installing-upgrading/product-lifecycle/).

It also updates CI to check against newer railses.

There is a mention about jruby in the readme, but no tests on CI. I would suggest just dropping it, as I am not sure many people (anyone?) uses it in production to worth the effort.


fatkodima:
I also noticed, that there is a `rubocop` and `standardrb`. I would suggest removing the latter and properly use `rubocop` and fix all the current offenses.

Also, there is a https://github.com/danmayer/coverband/blob/204d2d8b8511baee178a1b8db4f562ef14e35f96/Rakefile#L11 task, which runs `forked_tests`. These are not ran on CI currently and are broken. That should be also fixed.

Gemfile.rails7.2
```diff
@@ -4,7 +4,7 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in coverband.gemspec
 gemspec
-gem 'rails', '~>7'
```
danmayer, line 7: we might want a rails7 rails 7.1 and rails 7.2

lib/coverband/adapters/hash_redis_store.rb
```diff
@@ -25,14 +25,14 @@ def initialize(redis, opts = {})
         @save_report_batch_size = opts[:save_report_batch_size] || 100
         @format_version = REDIS_STORAGE_FORMAT_VERSION
         @redis = redis
-        raise "HashRedisStore requires redis >= 2.6.0" unless supported?
+        raise "HashRedisStore requires redis >= 6.0.0" unless supported?
```
danmayer, line 28: Do we have a reason to require redis 6, we were checking this based on features that redis supports but we aren't using any of the more modern and newer redis features.

lib/coverband/collectors/route_tracker.rb
```diff
@@ -13,7 +13,7 @@ class RouteTracker < AbstractTracker
       TITLE = "Routes"
 
       def initialize(options = {})
-        if Rails&.respond_to?(:version) && Gem::Version.new(Rails.version) >= Gem::Version.new("6.0.0") && Gem::Version.new(Rails.version) < Gem::Version.new("7.1.0")
+        if Rails&.respond_to?(:version) && Gem::Version.new(Rails.version) < Gem::Version.new("7.1.0")
           require_relative "../utils/rails6_ext"
```
danmayer, line 17: in this case we can delete this check and delete the rails6_ext file

danmayer:
Thanks... I guess we could start putting together a major release.

* I am good with updating to all the latest supported versions
* The JRuby tests were flaky and causing issues when I moved to github actions, but the support was added due to some requests, so I don't want to drop it. We can either fix the tests or add a comment that it is no longer tested
* I will try to fix the forked tests on CI they worked with a small tweak on my machine.

on Rubocop and Standard, I prefer standardrb for all my projects, I have rubocop as some editor tool chains seem to struggle with using standard as a auto formatter. Rubocop is supposed to use standard via the config file but appears not to do so...

* If we are going to remove one, I would remove all the references to rubocop
* CI enforces standardrb passing
* standard is currently passing
```
danmayer@dellstation:~/projects/coverband$ bundle exec standardrb
Inspecting 153 files
.........................................................................................................................................................
```

danmayer:
I fixed the forked tests, I also fixed up the benchmark tasks.... those are all now running on CI.

See the few comments I left after that I am happy to merge this.

lib/coverband/collectors/route_tracker.rb
fatkodima: We still support rails 7.0, but the feature was added in 7.1 (https://github.com/rails/rails/pull/43755)

fatkodima (commented):
Updated PR with removing old dummy apps from `test/` directory.

fatkodima:
I honestly prefer rubocop over standardrb, because I like full control and also when trying to ignore something with standardrb it starts looking like a hack for me. So I suggest having only one of them. If you want to remove one of them, please do, or let me know and I can do this.

danmayer:
> There is no reason, only as encouraging people upgrading their redis. I also doubt that people are on the latest version of coverband, but on the redis which is EOL many years ago. But happy to revert.

yeah let's leave it... Many users of Coverband are trying to use it because they are maintaining old legacy apps, so we often have users that are way behind the curve in terms of updates. You can see various issues people open wanting support for older Ruby and Rails versions... I want to be able to update but I do try to keep a pretty wide support window for that reason.

> I honestly prefer rubocop over standardrb...

yeah I understand on rubocop vs standardrb, everyone has some favorites there, I am more on the standardrb side of things as I don't want to tweak things to much...  but I can try to totally remove rubocop as it seems like running Rubocop with the standardrb mode doesn't really work correctly. I will take a look at cleaning that up tonight.

fatkodima:
Reverted redis changes.

danmayer (approved):

danmayer:
Thanks for all of this clean up, merged this main branch will be part of a major version release.

PR #567, opened 2024-11-15
Do not mark single-line and empty methods as dead

fatkodima:
Coverband is currently annoyingly marks single-line or empty methods as dead. I think, we can safely just ignore these.


danmayer:
thanks this looks good

PR #572, opened 2025-01-25
add BackgroundForMultiProcess

FeLvi-zzz:
## background
our production servers process a lot of resque jobs.

Coverband::ResqueWorker#perform reports when the job finished, so CPU utilization of our Redis store for collection coverage data goes too high.

## what I did
before sending reports to redis, an resque job saves coverage data to file(this should be on tmpfs).
background collector searches, collects and merges those files, and send to redis periodically.

## TODO before opening PR
- [ ] add spec
- [ ] test it in our staging/production env
- [ ] make it opt in

### DIFF
--- a/lib/coverband.rb
+++ b/lib/coverband.rb
@@ -25,6 +25,7 @@ require "coverband/reporters/base"
 require "coverband/reporters/console_report"
 require "coverband/integrations/background"
 require "coverband/integrations/background_middleware"
+require "coverband/integrations/background_for_multi_process"
 require "coverband/integrations/rack_server_check"
 require "coverband/configuration"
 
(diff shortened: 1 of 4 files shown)


danmayer:
I guess if this is a problem you have hit I could accept this PR, but make it opt-in so it would be off by default for resque users and can optionally be added when folks need the additional complexity.

FeLvi-zzz:
Yes, I guess almost all users don't need this function; it should be opt-in. In fact, we have already installed Coverband in some services, and they're not having any issues. I'm only having trouble with our largest service.
I'll make the modifications before opening the PR!

danmayer:
thanks sounds good just tag me again when ready for a full review

danmayer:
you want to wrap this up or anything else you want to make this not a draft and get it merged @FeLvi-zzz

FeLvi-zzz:
@danmayer 
sorry for being late... 
In conclusion, I adopted an another method instead of coverband for our largest service, so I have not need this PR.

I don't verify this feature in the service whether it is effective or not, I'll close this PR.

If other user need a feature like this, I'd like them to re-open and verify.

PR #577, opened 2025-01-31
Make minor improvements to GitHub actions

ydah:
Make minor improvements to GitHub actions

- Update GitHub Actions workflow to use GitHub format for standardrb
  
image
- Ensure that the standardrb job is performed only once
  - This should only need to be done once, not for each matrix.
- Use rake test:all instead of `bundle exec rake bundle exec rake forked_tests bundle exec rake benchmarks bundle exec rake benchmarks:memory`


danmayer (approved):
thanks

PR #591, opened 2025-08-22
Support views located in components directory, as well as sub-directories like packs/app/views

dkniffin:


danmayer (approved):

olivier-thatch:
@dkniffin I tried pulling the gem from GitHub to test this feature. Unfortunately, it doesn't seem to work: while Coverband now detects the views, it seems unable to detect when they are being used.

I did some limited testing, and it looks like that ViewComponent doesn't emit any `render_*.action_view` event. It can be optionally configured to emit its own `render.view_component` events (cf. [instrumentation docs](https://viewcomponent.org/guide/instrumentation.html)), but that's no silver bullet as the event's payload doesn't reference the view file (only the Ruby source file for the component).

Were you able to get this to work correctly on your end? I wonder if maybe it works with ViewComponent 3.x or earlier, but not with the recently released 4.0 version.

dkniffin:
I'm currently using view_component 3.23.2. I did get this deployed into our prod environment, and I think I'm seeing the same behavior you described. I'm pretty sure I'm seeing all my components and partials and views now (before, anything in `packs` was missing, which is basically all views in my codebase). But like you, I don't see any components in the "Used Views" section.

I also just noticed the ruby files for the components mostly don't show up in the main Coverage tab. A small handful do, but many do not. I'm not sure what that's about. It might just be my own config.

olivier-thatch:
Alright, thanks for confirming! Since this seems to be a limitation of ViewComponent, I've opened a feature request here: https://github.com/ViewComponent/view_component/issues/2451

> I also just noticed the ruby files for the components mostly don't show up in the main Coverage tab. A small handful do, but many do not. I'm not sure what that's about. It might just be my own config.

AFAICT the Ruby files for our ViewComponents are properly detected by Coverband, so this is probably an issue with your setup.

danmayer:
yeah for now we are very tied to the ActiveSupport Notifications emitted by default for rails views if they match those events it should work, or if they publish a different event with all the required data we could add a listener for that.

I will hold publishing the gem until we have a confirmation of some way to make this work or I revert the merge. Thanks for all the posts and let me know when you all think we can move forward

Issue #634, opened 2026-04-01
Coverband Web (6.1.8) is broken on Rack 3.2.6


Jack12816:
Hey there 👋 First of all, thank you to you and all contributors for this great project and for the ongoing maintenance efforts!

Today [Rack 3.2.6 was released](https://github.com/rack/rack/compare/v3.2.5...v3.2.6) and this broke coverband web with the following error:

```
NoMethodError: undefined method 'end_with?' for an instance of Regexp (NoMethodError)

        @urls = @urls.map { |url| [url, url.end_with?('/') ? url : "#{url}/".freeze].freeze }.freeze
                                           ^^^^^^^^^^
/usr/local/bundle/gems/rack-3.2.6/lib/rack/static.rb:97:in 'block in Rack::Static#initialize'
/usr/local/bundle/gems/rack-3.2.6/lib/rack/static.rb:97:in 'Array#map'
/usr/local/bundle/gems/rack-3.2.6/lib/rack/static.rb:97:in 'Rack::Static#initialize'
/usr/local/bundle/gems/coverband-6.1.8/lib/coverband/reporters/web.rb:35:in 'Class#new'
/usr/local/bundle/gems/coverband-6.1.8/lib/coverband/reporters/web.rb:35:in 'Coverband::Reporters::Web#init_web'
/usr/local/bundle/gems/coverband-6.1.8/lib/coverband.rb:169:in 'Coverband::Reporters::Web#initialize'
```

On the Rack side [a new check](https://github.com/rack/rack/commit/6d434bbccdc2ffc3086dbce3930c3816945a35ec) was introduced to mitigate a path traversal vulnerability. On coverband the `Rack::Static` application is [configured with an array of RegExp's](https://github.com/danmayer/coverband/blob/main/lib/coverband/reporters/web.rb#L37). [Based on the docs](https://github.com/rack/rack/blob/v3.2.6/lib/rack/static.rb#L9-L91), I'm unsure if RegExp's were supported in the past by accident or by design, so it might or not be an upstream issue based on:

```diff
 def route_file(path)
-  @urls.kind_of?(Array) && @urls.any? { |url| path.index(url) == 0 }
+  @urls.kind_of?(Array) && @urls.any? { |url, url_slash| path == url || path.start_with?(url_slash) }
 end
```

`String#index` allowed to use RegExp's as pattern.

[...]

PR #635, opened 2026-04-01
Fix: use empty string in Rack::Static urls instead of incorrect usage

alpaca-tc:
The usage of providing non-string values in the urls array was incorrect. related: 
We need to merge this PR before we can use a version that includes the Rack security patch. https://github.com/rack/rack/commit/6d434bbccdc2ffc3086dbce3930c3816945a35ec

`Rack::Static` expects urls to be an array of strings, but coverband was using it incorrectly.

### DIFF
--- a/lib/coverband/reporters/web.rb
+++ b/lib/coverband/reporters/web.rb
@@ -32,9 +32,9 @@ module Coverband
 
       def init_web
         full_path = Gem::Specification.find_by_name("coverband").full_gem_path
-        @static = Rack::Static.new(self,
-          root: File.expand_path("public", full_path),
-          urls: [/.*\.css/, /.*\.js/, /.*\.gif/, /.*\.png/])
+        @file_server = Rack::Files.new(
+          File.expand_path("public", full_path)
+        )
       end
 
       def check_auth
@@ -85,7 +85,7 @@ module Coverband
           else
             case request_path_info
             when /.*\.(css|js|gif|png)/
-              @static.call(env)
+              @file_server.get(env)
             when %r{/settings}
               [200, coverband_headers, [settings]]
             when %r{/view_tracker_data}
--- a/test/coverband/reporters/web_test.rb
+++ b/test/coverband/reporters/web_test.rb
@@ -45,6 +45,16 @@ module Coverband
       get "/json?line_coverage=true"
       assert last_response.ok?
     end
+
+    test "renders static files" do
+      get "/application.js"
+      assert last_response.ok?
+    end
+
+    test "renders 404 if static file doesn't exist" do
+      get "/unknown.js"
+      assert last_response.not_found?
+    end
   end
 end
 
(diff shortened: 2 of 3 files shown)


alpaca-tc:
@danmayer It was also confirmed by the Rack maintainers that Regexp is not an acceptable argument.
https://github.com/rack/rack/issues/2441

lib/coverband/reporters/web.rb
```diff
@@ -32,9 +32,7 @@ class Web
 
       def init_web
         full_path = Gem::Specification.find_by_name("coverband").full_gem_path
-        @static = Rack::Static.new(self,
-          root: File.expand_path("public", full_path),
-          urls: [/.*\.css/, /.*\.js/, /.*\.gif/, /.*\.png/])
+        @static = Rack::Static.new(self, root: File.expand_path("public", full_path), urls: [""])
```
ioquatix:
```suggestion
        @static = Rack::Static.new(self, root: File.expand_path("public", full_path))
```

Is this also okay?

ioquatix:
Hmm, those failures look weird.

alpaca-tc:
@ioquatix In the original implementation, it used `Rack::Static.new(self, ...)`, which caused it to fall back to itself and resulted in an infinite loop. Since the intended use case here is to serve files, I replaced it with `Rack::Files`.
I've force-pushed the changes 😄

zzak:
Ran into this too, the patch works for me 👍 Thanks!

This was also reported in #634.

Maybe squash the commits?

alpaca-tc:
@zzak Thanks for the feedback!

The commit 9a1a78a80f58a4efe7992a59bec1361f9118b039 is a fix for the CI on the main branch and is not directly related to the changes in this PR.
So I think it might not be necessary to squash the commits in this case.

That said, if there's a preference or guideline like "one PR = one commit," I'm happy to split it into a separate PR-just let me know 🙏

severin:
@danmayer FYI: this broke compatibility with _rack_ v1 (`Rack::Files` only exists in _rack_ v3, it was called `Rack::File` previously).

I know of course that _rack_ v1 is very outdated, but there are legacy apps out there still running on it 😉 Coverband does not explicitly depend on _rack_ and I'm not sure if you want to maintain compatibility with old _rack_ versions, but I feel that it would be helpful to call out _rack_ compatibility in the README.

2called-chaos:
Since we can't upgrade coverband atm due to legacy rails I did this (didn't want to monkeypatch the big call method so I kept the @static name and patched the call method on top). In case another one is in legacy hell

```ruby
class FixedCoverbandReporter < Coverband::Reporters::Web
  def init_web
    full_path = Gem::Specification.find_by_name("coverband").full_gem_path
    @static = Rack::Files.new(File.expand_path("public", full_path))

    def @static.call(env)
      get(env)
    end
  end
end
```

PR #640, opened 2026-04-17
Fix concrete_target glob to match packwerk pack directories

dkniffin:
Commit 329fa632 changed the ViewTracker glob from `**` to explicit paths `{,packs,engines/*}` to avoid traversing the entire project tree. However, `packs` only matches `packs/app/views/...`, not the standard packwerk structure of `packs/*/app/views/...`, causing concrete_target to return 0 results for packwerk-based apps.


danmayer (approved):
THANKS

Issue #665, opened 2026-09-10
Dark theme for the web report


naofumi-fujii:
**Is your feature request related to a problem? Please describe.**

The web report is light-only. With the dead code review session from #647 it is somewhere you sit for a while scrolling source files, and on a dark desktop it is the one window that stays bright.

`roadmap.md` already lists `full theming` under "improved web reporting", so this may be yours to do — asking before writing any code.

**Describe the solution you'd like**

A dark theme driven by `prefers-color-scheme`: no config, no toggle UI.

The blocker is the shape of `public/application.css` — 896 lines, **0** custom properties, 170 hard-coded hex colors, plus 17 `url(images/ui-*.png)` refs whose jQuery UI chrome cannot be recolored from CSS at all.

Two PRs:

1. **Refactor only** — lift the existing colors into custom properties on `:root`. Rendered output identical; useful on its own to anyone restyling the report.
2. **Dark theme** — a `@media (prefers-color-scheme: dark)` block redefining them. `.source_table .covered` / `.missed` / `.never` / `.skipped` picked for contrast rather than darkened mechanically; sprites start with `filter: invert()` on `.ui-icon`.

**Describe alternatives you've considered**

- A separate `application-dark.css` — duplicates 896 lines and guarantees drift.
- A config option (`config.web_theme = :dark`) — more surface to document and test, and it cannot follow the OS. Happy to build it this way if you want the choice per-deployment rather than per-viewer.
- Browser extensions — per-user, and they mangle the coverage row colors, which are the part that carries meaning.

**Additional context**

- Is now a bad time? `main` is at `7.0.0` with no tag yet. Happy to hold until it ships, or to send step 1 now since it touches nothing the 7.0.0 work does.
- `prefers-color-scheme`, or configurable?

If this sounds reasonable I will open the PRs — and if you would rather do the theming yourself, feel free to close.

danmayer:
yeah if you want to send a PR to add dark theme support I will take a look and merge it. I am working on the larger 7.0.0 release but I had a few more things I was going to do before I went live so dark mode could join that release. Send the PRs my way @naofumi-fujii  I appreciate it.

naofumi-fujii:
Thanks — both PRs are up.

- #666 — refactor only, lifts the colors into custom properties. Expanding every `var()` back reproduces the file byte for byte, so nothing renders differently.
- #667 — the `prefers-color-scheme` dark theme. Stacked on #666, so its diff will shrink to one commit once that merges.

[...]

PR #667, opened 2026-09-11
Add a dark theme to the web report

naofumi-fujii:
Second of the two PRs from #665.

**Stacked on #666** — it carries that commit too, so the reviewable diff here is the second commit ([9ab99ee](https://github.com/danmayer/coverband/pull/667/commits)), a single block appended to `public/application.css`. GitHub will drop #666's commit from this diff once it merges.

Driven by `prefers-color-scheme`, so it follows the OS with nothing to configure and no toggle to maintain.

The block redefines 137 of the 138 custom properties from #666 (the ColorBox overlay is already black). Coverage row colors are picked for contrast against the dark surface rather than darkened mechanically, so covered / missed / never / skipped stay as easy to tell apart as they are today.

## The two things a color alone cannot fix

- **jQuery UI widget chrome** paints an opaque light gradient PNG *over* the background color, so darkening the color does nothing — the sort header would stay white. Those `background-image`s are dropped in dark mode.
- **The icon sprites** are flat PNGs in a fixed color. Every neutral state now points at the darkest sheet with `filter: invert(1)`, which lands them all on one light gray and keeps the table sort arrows legible. The highlight and error sheets are already saturated enough to read on a dark surface and are left alone.

`magnify.png` is left as is — mid-gray on transparent, it still reads on the dark card. Happy to invert it via a pseudo-element if you would rather it match.

## Verification

Every foreground/background pair clears WCAG AA (4.5:1) in dark mode, and several are better than they are today:

| | light | dark |
|---|---|---|
| green status text | 3.78 | 6.53 |
| yellow status text | 2.14 | 6.57 |
| improve button label | 2.78 | 5.13 |
| card body text | 15.91 | 10.75 |
| source text on coverage rows | 13.26–20.31 | 11.01–14.31 |
| sortable table header | 5.97 | 7.55 |

The exception is the low-emphasis syntax colors (comment, preprocessor, chunk) at 4.23:1 — below AA, but well up from 1.47–1.82:1 in the light theme, where they are currently near-invisible on the pastel coverage rows.

Checked in a browser in both schemes: file list, sort headers and arrows, source view with all four row states, and the dead-code review session toolbar.


danmayer:
this is good and thanks a couple things I think are worth fixing quickly before merging

Two things I'd like addressed before merge, then I'm happy to take it.

### 1. Scope the media query to `screen`

The report stylesheet is linked for print as well as screen — `views/layout.erb:9`,
`views/abstract_tracker.erb:8`, `views/settings.erb:7` all use:

    <link href='...application.css' media='screen, projection, print' ...>

Chrome keeps reporting `prefers-color-scheme: dark` while printing, and background
colors are dropped from print output by default. So a user on a dark OS who prints or
PDFs a coverage report gets `#c9d1d9` text on white — close to blank. That print media
type is deliberate here, so I don't think this is hypothetical.

One word fixes it:

```css
@media screen and (prefers-color-scheme: dark) {

2. The highlight/error icon sheets aren't actually left alone

The comment above the .ui-icon block (line ~1248) says the saturated sheets are
untouched, but source order overrides them. The light rules are:

.ui-state-highlight .ui-icon { ...ui-icons_2e83ff...; }   /* specificity 0,2,0 */
.ui-state-error .ui-icon     { ...ui-icons_cd0a0a...; }   /* specificity 0,2,0 */

Your selector list includes .ui-widget-content .ui-icon and .ui-widget-header .ui-icon, which are also 0,2,0 and come later in the file. Since
.ui-state-highlight / .ui-state-errorget body, those
icons tie on specificity, lose to source order, and end up on the 222222 sheet with
filter: invert(1) applied.

In practice nothing regresses today —e sort arrows) is the
only .ui-icon this report actually renders — so this is latent rather than visible.
But I'd rather the code match its comt next. Either narrow
the two widget selectors:

  .ui-widget-content .ui-icon:not(.ui-state-highlight .ui-icon),

...or, simpler and easier to read, just re-assert the two sheets after the block:

  .ui-state-highlight .ui-icon {
    background-image: url(images/ui-i
    filter: none;
  }
  .ui-state-error .ui-icon,
  .ui-state-error-text .ui-icon {
    background-image: url(images/ui-icons_cd0a0a_256x240.png);
    filter: none;
  }

